PR #20, opened 2020-05-09
use CDK to deploy the lambda service

vincentsarago:
This is a small refactor to handler deploying the AWS Lambda using CDK. 

Deployement is working great but there seems to be an issue in the application when trying to get the `host` schema in https://github.com/developmentseed/titiler/blob/master/titiler/api/api_v1/endpoints/metadata.py#L59

I've opened an issue over https://github.com/erm/mangum/issues/105

Thanks to @leothomas and @ciaranevans who helped on this task


stack/app.py
```diff
@@ -4,17 +4,96 @@
...
+            reserved_concurrent_executions=concurrent,
+            timeout=core.Duration.seconds(timeout),
+        )
+
+        # defines an API Gateway Http API resource backed by our "dynamoLambda" function.
+        apigw.HttpApi(
+            self,
+            f"{id}-endpoint",
+            default_integration=apigw.LambdaProxyIntegration(handler=lambda_function),
+        )
```
vincentsarago, line 75: This just landed in cdk 🎉

PR #22, opened 2020-05-13
Add default env to lambda env

drewbo:


stack/app.py
```diff
@@ -65,6 +65,7 @@ def __init__(
             memory_size=memory,
             reserved_concurrent_executions=concurrent,
             timeout=core.Duration.seconds(timeout),
+            environment=DEFAULT_ENV,
```
vincentsarago, line 68: @drewbo can we do something similar like for ECS

```
        lambda_env = DEFAULT_ENV.copy()
        lambda_env.update(env)

        lambda_function = aws_lambda.Function(
            self,
            f"{id}-lambda",
            runtime=aws_lambda.Runtime.PYTHON_3_7,
            code=self.create_package(code_dir),
            handler="handler.handler",
            memory_size=memory,
            reserved_concurrent_executions=concurrent,
            timeout=core.Duration.seconds(timeout),
            environment=lambda_env,
        )

```

PR #24, opened 2020-05-28
add custom colormaps

vincentsarago:
This PR refactor how tiles endpoint parameters are defined and use dependencies to register custom colormaps


titiler/api/deps.py
```diff
@@ -0,0 +1,42 @@
...
+from typing import Union, Optional
+
+import re
+from fastapi import Query
+import numpy
+
+from titiler.custom import cmap as custom_colormap
+from rio_tiler.colormap import cmap
+
+cmap.register("above", custom_colormap.above_cmap)
```
vincentsarago, line 12: Here we register a custom colormap.

vincentsarago:
Question to answer: 

Registering custom colormap needs to happen only once (at app startup) or rio-tiler will raise an error (you cannot re-register the same cmap).

Is `deps` the best way to do: maybe not ;-)

PR #39, opened 2020-06-25
add width,height and resampling params options

vincentsarago:


titiler/api/deps.py
```diff
@@ -127,14 +136,21 @@ def __init__(
             title="Color Formula",
             description="rio-color formula (info: https://github.com/mapbox/rio-color)",
         ),
-        color_map: Optional[ColorMapName] = Query(
+        color_map: Optional[ColorMapNames] = Query(
             None, description="rio-tiler's colormap name"
         ),
+        resampling_method: ResamplingNames = Query(
+            ResamplingNames.nearest, description="Resampling method."  # type: ignore
+        ),
```
vincentsarago, line 144: This is just for documentation and validation. 

`resampling_method` will be passed via `kwargs`

Issue #3, opened 2019-07-12
add mosaics


vincentsarago:

vincentsarago:
## Mosaic Store 

Following some discussion on slack with @geospatial-jeff we need to choose: 
- what backend we support (file, s3, http, dynamoDB, stac) 
- how the user pass the mosaicjson (url, id) 
- do we allow external mosaic ? 


in cogeo-mosaic-tiler, the mosaicjson is passed either via `url` or via a `mosaicid` (prefered method). With the mosaicid we know (app owner) that the mosaic will be store in `s3://${mosaicbuck}/${mosaicprefix}/{mosaicid}.json.gz` so we added endpoint like `/tiles/{mosaicid}/z/x/y`

The mosaicid is created using the sha224 of the mosaic body. 

here is how we reconstruct the path (the datastore is S3 and is defined via env variable)
https://github.com/developmentseed/cogeo-mosaic-tiler/blob/master/cogeo_mosaic_tiler/handlers/app.py#L60-L67

#### IMO

- what backend we support (file, s3, http, dynamoDB, stac) 

For now S3 / File (EFS) / DynamoDB is fine, but titiler SHOULD have a default backend defined by a setting (e.g `s3://bucket/mosaics/` or `dynamodb://`

- how the user pass the mosaicjson (url, id) 

as for cogeo-mosaic-tiler, I think uniq ID for a mosaic is fine. That's said I'd love to use something more intelligible.

in #40 I throw the idea of using id like mapbox tileset: `{user}.{layername}`. This will also be good for a long term goal if we add `users` to titiler (like user have some COGs on our bucket, they can create mosaic ... ) 

- do we allow external mosaic ? 

I'll tend to answer NO, but a user could post a mosaic to add it to our system.

geospatial-jeff:
> For now S3 / File (EFS) / DynamoDB is fine, but titiler SHOULD have a default backend defined by a setting (e.g `s3://bucket/mosaics/` or `dynamodb://`

I think it should support all backends and raise `400 Bad Request` status code when trying to perform an unsupported operation (ex. write to HTTP backend).  I do like the idea of defining a default backend so you can do `id` lookups to the default backend and `url` lookups to any backend.  Could also encode the name of the backend into the id and not use `url` lookups at all.


> in #40 I throw the idea of using id like mapbox tileset: `{user}.{layername}`. This will also be good for a long term goal if we add `users` to titiler (like user have some COGs on our bucket, they can create mosaic ... )

[...]

PR #40, opened 2020-06-26
[WIP] mosaics

geospatial-jeff:
Add support for mosaicjson, will update the pr description later

Ref #3 

TODOs:
- [x] Add mosaicjson metadata routes.
- [x] Add tile routes.
- [x] Add tilejson route.
- [x] Add wmts route.
- [x] Add point route.
- [ ]  ~Add crop (bbox) route.~
- [x] Figure out and implement app-level caching strategy (cogeo-mosaic already does some LRU caching itself).
- [x] Finalize mosaic lookup (id/url, users etc.)
- [x] Add server timings to response headers
- [ ]  ~Add HTML viewer~
- [x] 95%+ test coverage of mosaic-related code


titiler/api/endpoints/cog.py
```diff
@@ -60,7 +60,7 @@ async def cog_bounds(
     response_model_exclude_none=True,
     responses={200: {"description": "Return basic info on COG."}},
 )
-async def cog_info(
+def cog_info(
```
geospatial-jeff, line 63: Switched to `def` so this can be reused in `/mosaicjson/info`.  These should all be normal `def` anyway since they aren't using coroutines.

titiler/models/metadata.py
```diff
@@ -22,14 +22,14 @@ class cogInfo(BaseModel):
...
     bounds: Tuple[float, float, float, float]
-    band_metadata: List[Tuple[int, Dict[int, Any]]]
     band_descriptions: List[Tuple[int, str]]
     dtype: str
     colorinterp: List[str]
     nodata_type: NodataTypes
     scale: Optional[float]
     offsets: Optional[float]
     colormap: Optional[Dict[int, ColorTuple]]
+    band_metadata: Optional[List[Tuple[int, Dict[int, Any]]]]
```
geospatial-jeff, line 32: Switched to optional to allow reuse for `/mosaicjson/info` response model -- I don't think band metadata makes sense for mosaicjson metadata.  Should maybe write a `mosaicInfo` model instead and revert this change.

titiler/api/endpoints/mosaic.py
```diff
@@ -11,10 +12,37 @@
...
 from fastapi import APIRouter, Query
+from fastapi.routing import APIRoute
 
+from starlette.exceptions import HTTPException
+from starlette.requests import Request
 from starlette.responses import Response
 
-router = APIRouter()
+
+class MosaicJSONRouter(APIRoute):
```
geospatial-jeff, line 22: Including stac backend complicates things for initial development.  Fully anticipate to remove this in the future.

titiler/api/endpoints/mosaic.py
```diff
@@ -126,3 +145,77 @@ def mosaicjson_info(
...
+    # executor provided by the event loop.  Instead of calling ``rio_tiler_mosaic.mosaic.mosaic_tiler`` directly we will
+    # transcribe the code here and use the executor provided by the event loop.  This also means we define this function
+    # as a coroutine (even though nothing that is called is a coroutine), since the event loop's executor isn't
+    # available in normal ``def`` functions.
+    # https://github.com/cogeotiff/rio-tiler-mosaic/blob/master/rio_tiler_mosaic/mosaic.py#L37-L102
+    _tiler = partial(cogeoTiler, tile_x=x, tile_y=y, tile_z=z, tilesize=tilesize)
+    futures = [run_in_threadpool(_tiler, asset) for asset in assets]
+
+    # TODO: parametrize concurrency
+    semaphore = asyncio.Semaphore(10)
```
geospatial-jeff, line 184: Semaphore is used to limit concurrency instead of [chunking list of assets](https://github.com/cogeotiff/rio-tiler-mosaic/blob/master/rio_tiler_mosaic/mosaic.py#L90-L91)

titiler/api/endpoints/mosaic.py
```diff
@@ -175,6 +177,9 @@ async def mosaic_tile(
 
     tilesize = 256 * scale
 
+    # TODO: Parametrize/support other tms
+    tms = morecantile.tms.get(TileMatrixSetNames.WebMercatorQuad.name)  # type: ignore
```
vincentsarago, line 181: I'm good supporting only WebMercator, supporting other tms will be hard

titiler/api/endpoints/mosaic.py
```diff
@@ -231,9 +236,10 @@ async def mosaic_tile(
     else:
         driver = drivers[format.value]
         options = img_profiles.get(driver.lower(), {})
-        # TODO: Support tif
         if format == ImageType.tif:
-            raise Exception("Please support me")
```
vincentsarago: ❤️

titiler/api/endpoints/mosaic.py
geospatial-jeff: Why do you say that?  Would it be hard mapping `rio_tiler_crs` across a list of assets compared to `rio_tiler`?

vincentsarago: I was thinking about the quadkey!

geospatial-jeff: I haven't looked at morecantile in a second, is there a feasible way of validating 2x decimation for a particular tile matrix set?  So that those which fit the quadkey tiling scheme may be supported and those that don't would not be supported.

Assuming that a tile matrix set has 2x decimation I think it is realistic to support here, please correct if that is wrong.

vincentsarago: well in fact I'm not sure if the TMS in the tiler should be == in the tms/quadkey used for the mosaic 🤷‍♂️ maybe it's time to think about another way to index the tiles

titiler/api/endpoints/cog.py
```diff
@@ -101,7 +101,7 @@ async def cog_metadata(
     return info
 
 
-params: Dict[str, Any] = {
+tile_response_codes: Dict[str, Any] = {
```
vincentsarago, line 104: Maybe we can moves this in https://github.com/developmentseed/titiler/blob/master/titiler/ressources/responses.py or https://github.com/developmentseed/titiler/blob/master/titiler/ressources/common.py

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,245 @@
...
+from starlette.concurrency import run_in_threadpool
+from starlette.exceptions import HTTPException
+from starlette.requests import Request
+from starlette.responses import Response
+
+
+def _chunks(my_list: Sequence, chuck_size: int):
+    """Yield successive n-sized chunks from l."""
+    for i in range(0, len(my_list), chuck_size):
+        yield my_list[i : i + chuck_size]
```
vincentsarago, line 41: can we use https://github.com/cogeotiff/rio-tiler/blob/cb0a709144b4589bb9940e56716eb741a53f71fd/rio_tiler/utils.py#L23-L26 ?

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,245 @@
...
+                url = body.get("url")
+            if url.startswith("stac"):
+                # Raise as a validation error
+                raise HTTPException(
+                    status_code=422,
+                    detail="STAC is not currently supported for mosaicjson endpoints",
+                )
+            return await original_route_handler(request)
+
+        return forbid_stac_backend
```
vincentsarago, line 64: I'm not really sure we need this, the app will just fail if you try `stac...` url

titiler/api/endpoints/mosaic.py
geospatial-jeff: This actually shouldn't be here (I forgot to delete), since we are using `asyncio.Semphore` instead of using `_chunks` to limit concurrency.  I actually didn't realize this helper function was defined in `rio_tiler`, if it is needed in the future projects I will import from there.

titiler/api/endpoints/cog.py
geospatial-jeff: Yes I agree it should be shared.  Although maybe having a `models/openapi.py` file or equivalent (maybe that lives in ressources?) makes the most sense here.  I would think that a lot of the openapi stuff is shared across routes.

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,245 @@
...
+)
+def create_mosaicjson(body: CreateMosaicJSON):
+    """Create a MosaicJSON"""
+    mosaic = MosaicJSON.from_urls(
+        body.files,
+        minzoom=body.minzoom,
+        maxzoom=body.maxzoom,
+        max_threads=body.max_threads,
+    )
+    with MosaicBackend(body.url, mosaic_def=mosaic) as mosaic:
```
vincentsarago, line 81: I'm kinda hesitant with letting user write where they want. In an ideal world the mosaic id should be something like `username.layer` (like mapbox tileset https://github.com/mapbox/pxm-manifest-specification/blob/master/schemas/pxm-manifest-0.5.1.json#L22)

vincentsarago: by default the app should either write to efs or to s3 IMO

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,245 @@
...
+    response_model=MosaicJSON,
+    response_model_exclude_none=True,
+    responses={200: {"description": "Return MosaicJSON definition"}},
+)
+def read_mosaicjson(
+    resp: Response, url: str = Query(..., description="MosaicJSON URL")
+):
+    """Read a MosaicJSON"""
+    resp.headers["Cache-Control"] = "max-age-3600"
+    with MosaicBackend(url) as mosaic:
```
vincentsarago, line 97: Maybe the same as earlier comment, url could be like a tileset name

vincentsarago: In an ideal world we will have user token and let user only see their mosaic

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+@router.get(
+    "",
+    response_model=MosaicJSON,
+    response_model_exclude_none=True,
+    responses={200: {"description": "Return MosaicJSON definition"}},
+)
+def read_mosaicjson(
+    resp: Response, url: str = Query(..., description="MosaicJSON URL")
```
vincentsarago, line 93: if we follow the `user.layer` schema, what do you think about adding this in the path ? 

e.g `endpoint/mosaicjson/kylebarron.somethingawesome`

titiler/api/endpoints/mosaic.py
geospatial-jeff: Off the top of my head those are the only backends which support write.  Whats missing (and will be added later) is backends which don't support `write` (i.e. raise `NotImplementedError`) will raise some sort of exception.

I've just been using S3 backend while developing because it supports basically everything.

titiler/api/endpoints/mosaic.py
geospatial-jeff: I wanted to be explicit here for now (rather than raise a `500` because we aren't handling **kwargs in `MosaicBackend`).  Besides that i agree with you.  Ideally STACBackend is fully supported in the near future for both COG/MosaicJSON and we can remove this completely.

titiler/api/endpoints/mosaic.py
geospatial-jeff: If we went with `user.layer` I think is the most "restful" way (`/endpoint/resource/<id>`).

The more I think about that only EFS and S3 support write methods the more I like this idea.

titiler/api/endpoints/mosaic.py
geospatial-jeff: For now I removed morecantile and am using `mercantile` to calculate tile bounds.

titiler/api/endpoints/mosaic.py
geospatial-jeff: Resolving in favor of https://github.com/developmentseed/titiler/issues/3

geospatial-jeff:
With the last commit all routes support both `mosaic_id` and `url` parameters (similar to cogeo-mosaic-tiler).  If neither are present in the call a `422` is raised.  I also added two environment variables for configuring default backend.
- `DEFAULT_MOSAIC_BACKEND` (defaults to `s3://`)
- `DEFAULT_MOSAIC_HOST`

If `mosaic_id` is provided titiler will look up the mosaic from the default backend/host.  If url is provided, titiler just passes the url to the `MosaicBackend` constructor which uses the url scheme to instantiate the appropriate backend.

setup.py
```diff
@@ -16,6 +16,7 @@
 extra_reqs = {
     "dev": ["pytest", "pytest-cov", "pytest-asyncio", "pre-commit"],
     "stac": ["stac-tiler~=0.0rc2"],
+    "mosaic": ["cogeo-mosaic==3.0a4", "rio-tiler-mosaic"],
```
vincentsarago, line 19: `rio-tiler-mosaic` is not needed here

geospatial-jeff: Removed with https://github.com/developmentseed/titiler/pull/40/commits/097927aaaa59b9a942d4a84e3e9ea31ef982a808

titiler/api/endpoints/mosaic.py
geospatial-jeff: Removed the custom router because it happens before dependency injection which messes with parameter validation.  I think it is fine for now to throw `500` for stac backend.

https://github.com/developmentseed/titiler/pull/40/commits/180a3d1be78e1bf325c22baf5894463c252ef221

geospatial-jeff:
@vincentsarago Do you think a preview route should be supported for mosaics?  I'm on the fence.

vincentsarago:
> @vincentsarago Do you think a preview route should be supported for mosaics?  I'm on the fence.

No I don’t think we need preview or crop. For another project I’ve got preview implemented using WMTS and fetching tiles instead of the files directly but I don’t think we need this by default in titiller

titiler/settings.py
```diff
@@ -0,0 +1,10 @@
+"""Application environment variables"""
+
+from starlette.config import Config
+
+config = Config()
+
+DEFAULT_MOSAIC_BACKEND = config("DEFAULT_MOSAIC_BACKEND", cast=str, default="s3://")
+
+# TODO: Remove default (default is currently set to pass tests)
+DEFAULT_MOSAIC_HOST = config("DEFAULT_MOSAIC_HOST", cast=str, default="placeholder")
```
vincentsarago, line 10: can't you use fake value in https://github.com/developmentseed/titiler/blob/master/tests/conftest.py#L22 for testing ?

vincentsarago: also those should only be required if we use the mosaic ?

titiler/api/deps.py
```diff
@@ -230,3 +232,21 @@ def __init__(
...
+        url: Optional[str] = Query(None, description="MosaicJSON URL"),
+    ):
+        """Create mosaic path from args"""
+        if not mosaic_id and not url:
+            raise HTTPException(
+                status_code=422, detail="Missing 'mosaic_id' or 'url' parameter"
+            )
+        self.mosaic_path = (
+            url or f"{DEFAULT_MOSAIC_BACKEND}{DEFAULT_MOSAIC_HOST}/{mosaic_id}.json.gz"
+        )
```
vincentsarago, line 252: so, if we don't use mosaicid in the path maybe we can use a url like parameter like `mosaicid://{mosaicid}`?

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,401 @@
...
+    tile_format: Optional[ImageType] = Query(
+        None, description="Output image type. Default is auto."
+    ),
+    mosaic_params: CommonMosaicParams = Depends(),
+):
+    """Create TileJSON"""
+    kwargs = {"z": "{z}", "x": "{x}", "y": "{y}", "scale": tile_scale}
+    if tile_format:
+        kwargs["format"] = tile_format
+    tile_url = request.url_for("mosaic_tile", **kwargs).replace("\\", "")
```
vincentsarago, line 161: 😱 so simple

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,401 @@
...
+                    tile, mask = await fut
+                except Exception:
+                    # Gracefully handle exceptions
+                    continue
+
+                tile = numpy.ma.array(tile)
+                tile.mask = mask == 0
+                pixsel.feed(tile)
+                if pixsel.is_done:
+                    break
```
vincentsarago, line 299: maybe it's time to re-write rio-tiler-mosaic using async?

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,401 @@
...
+    # available in normal ``def`` functions.
+    # https://github.com/cogeotiff/rio-tiler-mosaic/blob/master/rio_tiler_mosaic/mosaic.py#L37-L102
+    _tiler = partial(
+        cogeoTiler,
+        tile_x=x,
+        tile_y=y,
+        tile_z=z,
+        tilesize=tilesize,
+        indexes=image_params.indexes,
+        # expression=image_params.expression, # TODO: Figure out why expression kwarg doesn't work
```
vincentsarago, line 278: it's not working because you are using rio-tiler and not rio-tiler-crs (which as a native handle for expression). This might change in near future but for now I'll advice to use rio-tiler-crs (as in the COG endpoint)

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,401 @@
...
+        tile = utils.postprocess(
+            tile,
+            mask,
+            rescale=image_params.rescale,
+            color_formula=image_params.color_formula,
+        )
+    timings.append(("Post-process", t.elapsed))
+
+    bounds = mercantile.xy_bounds(mercantile.Tile(x, y, z))
+    dst_transform = from_bounds(*bounds, tilesize, tilesize)
```
vincentsarago, line 319: both rio-tiler and rio-tiler-crs have a function for this:
https://github.com/cogeotiff/rio-tiler/blob/cb0a709144b4589bb9940e56716eb741a53f71fd/rio_tiler/utils.py#L315
https://github.com/cogeotiff/rio-tiler-crs/blob/master/rio_tiler_crs/reader.py#L24-L34

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,401 @@
...
+    ),
+    tile_scale: int = Query(
+        1, gt=0, lt=4, description="Tile size scale. 1=256x256, 2=512x512..."
+    ),
+    mosaic_params: CommonMosaicParams = Depends(),
+):
+    """OGC WMTS endpoint."""
+    scheme = request.url.scheme
+    host = request.headers["host"]
+    endpoint = f"{scheme}://{host}/mosaic"
```
vincentsarago, line 356: maybe we can use `url_for("read_mosaicjson")`?

titiler/settings.py
```diff
@@ -0,0 +1,10 @@
+"""Application environment variables"""
+
+from starlette.config import Config
+
+config = Config()
```
vincentsarago, line 5: I know this is not part of the scope of this PR but it will be great to unify `settings` and `titiler.core.config`

titiler/api/deps.py
geospatial-jeff: Thats a fantastic idea

titiler/settings.py
geospatial-jeff: Agreed, I think I will move the contents of `settings` into `titiler.core.config`.

titiler/api/endpoints/mosaic.py
geospatial-jeff: Maybe.  I'd want to see a comparison between threadpool executor and asyncio first.  Asyncio is certainly better for FastAPI but may differ for lambda etc.

titiler/settings.py
vincentsarago: or keep `titiler.settings` only? up to you ;-)

geospatial-jeff: I think `settings` should be used in the future especially for environment vars but I think this PR to stick to mosaics and follow the current repo structure (ex. I've been trying to avoid deduping code in other routes even though it is very tempting)

titiler/api/endpoints/mosaic.py
geospatial-jeff: Oh I remember why I didn't use `url_for`, looks like the wmts template takes the base url of the router.

titiler/settings.py
geospatial-jeff: https://github.com/developmentseed/titiler/pull/40/commits/60ede4cb7bafa1b6b4e748dc784cb987493296cd

titiler/api/endpoints/mosaic.py
geospatial-jeff: https://github.com/developmentseed/titiler/pull/40/commits/52b416d61244e3cb02ea7b68cad252a3387de2b2

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,401 @@
...
+    semaphore = asyncio.Semaphore(int(os.getenv("MOSAIC_CONCURRENCY", 10)))
+
+    values = []
+    with utils.Timer() as t:
+        async with semaphore:
+            for fut in asyncio.as_completed(futures):
+                try:
+                    values.append(await fut)
+                except Exception:
+                    continue
```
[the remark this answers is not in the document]
geospatial-jeff: Is this what you meant? https://github.com/developmentseed/titiler/pull/40/commits/600c0d89192de827530f934e6f2d9f745de658b0

titiler/api/deps.py
geospatial-jeff: https://github.com/developmentseed/titiler/pull/40/commits/37f87415d21994e853230fb95f0e18771f8cbadc

titiler/api/endpoints/mosaic.py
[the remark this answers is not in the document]
vincentsarago: not really. I was hopping we could make a helper function that takes all the async block https://github.com/developmentseed/titiler/blob/600c0d89192de827530f934e6f2d9f745de658b0/titiler/api/endpoints/mosaic.py#L205-L222 outside

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,399 @@
...
+from starlette.templating import Jinja2Templates
+
+router = APIRouter()
+templates = Jinja2Templates(directory="titiler/templates")
+
+
+def _read_point(asset: str, *args, **kwargs) -> List:
+    """Read pixel value at a point from an asset"""
+    with rasterio.open(asset) as src_dst:
+        return point(src_dst, *args, **kwargs)
```
vincentsarago, line 50: Would you mind using rio-tiler-crs?

https://github.com/developmentseed/titiler/blob/725da5fa1bc56d8e192ae8ff0ad107493ca93378/titiler/api/endpoints/cog.py#L369-L370

titiler/api/endpoints/mosaic.py
vincentsarago: FYI using rio-tiler-crs you'll need to use a wrapper because of the contextManager

```python
with COGReader(asset) as cog:
    data, mask = cog.tile(x, y, z)
```

titiler/api/endpoints/mosaic.py
vincentsarago: https://github.com/arturo-ai/titiler/blob/52b416d61244e3cb02ea7b68cad252a3387de2b2/titiler/api/endpoints/mosaic.py#L76

`url_for("read_mosaicjson")` should return the base path `:endpoint:/mosaic` ?

geospatial-jeff: you're very right haha

titiler/api/endpoints/mosaic.py
geospatial-jeff: yeah i'll move everything over to rio-tiler-crs tomorrow

titiler/api/endpoints/mosaic.py
geospatial-jeff: Yeah I'll probably add a few more wrappers to global scope for now.  Ideally when we dedup code we can use dependency injection to wrap mosaic logic around the `with COGReader` block used by the cog routes.

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,399 @@
...
+    responses={200: {"description": "Return the bounds of the MosaicJSON"}},
+)
+def mosaicjson_bounds(resp: Response, mosaic_params: CommonMosaicParams = Depends()):
+    """Read MosaicJSON bounds"""
+    resp.headers["Cache-Control"] = "max-age=3600"
+    with MosaicBackend(mosaic_params.mosaic_path) as mosaic:
+        return {"bounds": mosaic.mosaic_def.bounds}
+
+
+@router.get("/info", response_model=cogInfo)
```
vincentsarago, line 115: It's a bit weird to use cogInfo for a mosaic metadata ?

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,399 @@
...
+        return response
+
+
+@router.get(
+    "/tilejson.json",
+    response_model=TileJSON,
+    responses={200: {"description": "Return a tilejson"}},
+    response_model_exclude_none=True,
+)
+async def mosaic_tilejson(
```
vincentsarago, line 149: I think we can remove the `async` here

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,399 @@
...
+        tjson = TileJSON(
+            name=mosaic.mosaic_def.name,
+            description=mosaic.mosaic_def.description,
+            attribution=mosaic.mosaic_def.attribution,
+            bounds=mosaic.mosaic_def.bounds,
+            center=mosaic.mosaic_def.center,
+            minzoom=mosaic.mosaic_def.minzoom,
+            maxzoom=mosaic.mosaic_def.maxzoom,
+            tiles=[tile_url],
+        )
```
vincentsarago, line 175: the mosaic backend has a `meta` propriety and I think we could do 

```
    with MosaicBackend(mosaic_params.mosaic_path) as mosaic:
         tjson = mosaic.meta.copy()
         tjson.update({"tiles": [tile_url]})
```

titiler/api/endpoints/mosaic.py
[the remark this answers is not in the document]
geospatial-jeff: I added an async generator to process coroutines https://github.com/developmentseed/titiler/pull/40/commits/7cf6ea9b104f6b8b238240bae6c22a11a5d62378

titiler/api/endpoints/mosaic.py
geospatial-jeff: Much better! https://github.com/developmentseed/titiler/pull/40/commits/38e14995c26202977313f4934d824962791d2394

titiler/api/endpoints/mosaic.py
geospatial-jeff: https://github.com/developmentseed/titiler/pull/40/commits/4602df3f26d527c3a5778a093f25d27e339040ff

titiler/api/endpoints/mosaic.py
geospatial-jeff: https://github.com/developmentseed/titiler/pull/40/commits/4602df3f26d527c3a5778a093f25d27e339040ff

Removed addition of cogInfo from mosaic metadata.  I think it is worth it to include some asset-level metadata about the mosaic like data type and photometric interpretation to name a few but this may be easier to implement after code dedup (ref #33).

Also this assumes that all assets in the MosaicJSON are the same, and I'm not sure if that assumption is always valid.

vincentsarago: yeah we can remove the `COG info` for now, IMO most of this should come from the mosaicJSON document ;-)

titiler/api/deps.py
```diff
@@ -230,3 +232,19 @@ def __init__(
...
+    """Common mosaic params."""
+
+    def __init__(
+        self, url: str = Query(None, description="MosaicJSON URL"),
+    ):
+        """Create mosaic path from args"""
+        parsed = urlparse(url)
+        if parsed.scheme == "mosaicid":
+            self.mosaic_path = (
+                f"{DEFAULT_MOSAIC_BACKEND}{DEFAULT_MOSAIC_HOST}/{parsed.scheme}.json.gz"
```
vincentsarago, line 247: I don't think it should be `parsed.scheme` here!

titiler/models/mosaic.py
```diff
@@ -0,0 +1,49 @@
...
+    """MosaicJSON info."""
+
+    bounds: Tuple[float, float, float, float]
+    band_descriptions: List[Tuple[int, str]]
+    dtype: str
+    colorinterp: List[str]
+    nodata_type: NodataTypes
+    scale: Optional[float]
+    offsets: Optional[float]
+    colormap: Optional[Dict[int, ColorTuple]]
```
vincentsarago, line 43: This need to be updated!

titiler/api/deps.py
vincentsarago: ping!

vincentsarago:
@geospatial-jeff do you think you'll have time to finish this ?

geospatial-jeff:
Yeah I'm going to finish it up tonight, haven't had much free time the past 2 weeks

geospatial-jeff:
@vincentsarago 

For now I've removed the app-level cache for mosaics for several reasons:
1. There is an LRU cache used by `cogeo-mosaic`.  Currently the cache is not configurable, ideally we would disable the LRU cache in favor of caching with the memcache layer -- memcache will cache across multiple instances of titiler within the same deployment while LRU cache is restricted to a single instance.
2. With some code dedup across COG/mosaic (ref #33) it will be possible to cache a single tile across both COG and mosaic requests.  Currently this would rely on creating the same `request_id` across multiple requests which isn't feasible given `cog_tile` and `mosaic_tile` take different arguments.

For these two reasons I propose adding the cache layer to the mosaic routes once (1) `cogeo-mosaic` cache is configurable and (2) #33 has been closed.  Happy to write up issues in the appropriate repos if you think this is a good idea.

I also thing we can move the HTML viewer to a future PR, its definitely helpful and a nice feature but not critical.

README.md
```diff
@@ -162,10 +163,11 @@ titiler/                         - titiler python module.
  ├── api/                        - general api directory.
  │   ├── api_                 - directory hosting version 1 of the API.
  │   │   ├── endpoints           - api routes.
+ │   │   │   ├── cog.py          - COG related endpoints.
```
vincentsarago, line 166: I'll add stac.py ;-)

tests/conftest.py
```diff
@@ -1,13 +1,18 @@
 """``pytest`` configuration."""
 
+import json
 import os
+from typing import Any, Dict
 
 import pytest
+from rasterio import MemoryFile
```
vincentsarago, line 8: I usually do `from rasterio.io import MemoryFile` but if it works 👌

tests/routes/test_mosaic.py
```diff
@@ -0,0 +1,242 @@
...
+            data["tiles"][qk] = [
+                os.path.join(os.path.dirname(fname), f) for f in data["tiles"][qk]
+            ]
+        return MosaicJSON(**data)
+
+    return _read
+
+
+def test_read_mosaic(app):
+    response = app.get("/mosaicjson", params={"url": MOSAICJSON_FILE})
```
vincentsarago, line 33: I though pre-commit was going to fails without docstring here 🤷‍♂️

tests/routes/test_mosaic.py
```diff
@@ -0,0 +1,242 @@
...
+    assert body["values"][0] == [9943, 9127, 9603]
+
+
+def test_tile(app):
+    mosaicjson = read_json_fixture(MOSAICJSON_FILE)
+    bounds = mosaicjson["bounds"]
+    tile = mercantile.tile(*mosaicjson["center"])
+    partial_tile = mercantile.tile(bounds[0], bounds[1], mosaicjson["center"][-1])
+
+    with patch.object(FileBackend, "_read", mosaic_read_factory(MOSAICJSON_FILE)):
```
vincentsarago, line 151: 👍

tests/routes/test_mosaic.py
```diff
@@ -0,0 +1,242 @@
...
+    body = response.json()
+    assert len(body["values"]) == 1
+    assert body["values"][0] == [9943, 9127, 9603]
+
+
+def test_tile(app):
+    mosaicjson = read_json_fixture(MOSAICJSON_FILE)
+    bounds = mosaicjson["bounds"]
+    tile = mercantile.tile(*mosaicjson["center"])
+    partial_tile = mercantile.tile(bounds[0], bounds[1], mosaicjson["center"][-1])
```
vincentsarago, line 149: `mosaicjson["center"][-1]` -> `mosaicjson["minzoom"]` ?

titiler/api/deps.py
```diff
@@ -230,3 +232,19 @@ def __init__(
...
+        self, url: str = Query(..., description="MosaicJSON URL"),
+    ):
+        """Create mosaic path from args"""
+        parsed = urlparse(url)
+        if parsed.scheme == "mosaicid":
+            self.mosaic_path = (
+                f"{DEFAULT_MOSAIC_BACKEND}{DEFAULT_MOSAIC_HOST}/{parsed.netloc}.json.gz"
+            )
+        else:
+            self.mosaic_path = url
```
vincentsarago, line 250: Looking at this, I don't think we need to use a class here, a simple function that returns a `mosaic_path` should be sufficient ?

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,399 @@
...
+    """Create an async generator which executes coroutines at specified concurrency and yields response as completed"""
+    semaphore = asyncio.Semaphore(concurrency)
+    async with semaphore:
+        for fut in asyncio.as_completed(futures):
+            yield fut
+
+
+@router.post(
+    "", response_model=MosaicJSON, response_model_exclude_none=True,
+)
```
vincentsarago, line 67: this could fit in one line

titiler/api/endpoints/mosaic.py
```diff
@@ -0,0 +1,399 @@
...
+    ),
+    mosaic_params: CommonMosaicParams = Depends(),
+):
+    """Create TileJSON"""
+    kwargs = {"z": "{z}", "x": "{x}", "y": "{y}", "scale": tile_scale}
+    if tile_format:
+        kwargs["format"] = tile_format
+    tile_url = request.url_for("mosaic_tile", **kwargs).replace("\\", "")
+    with MosaicBackend(mosaic_params.mosaic_path) as mosaic:
+        tjson = TileJSON(**mosaic.metadata, tiles=[tile_url],)
```
vincentsarago, line 168: we don't need the trailing `,` here

geospatial-jeff: pre-commit letting me down

PR #48, opened 2020-07-14
Edits mosaics

vincentsarago:


vincentsarago:
@geospatial-jeff let me know what you think! we could still revisit this later

geospatial-jeff:
looks good to me!

Issue #50, opened 2020-07-14
switch to starlette config and use .env


vincentsarago:

vincentsarago:
closed in #55

PR #55, opened 2020-07-28
use starlette config

geospatial-jeff:
Ref #50 

- Remove `titiler.core` in favor of starlette config.

Haven't added an `.env` file yet, there are already 2 sets of environment variables used for aws sdk deployment and docker-compose and I'm hesitant to add a third.  We could use multiple env files (`.env.docker`, `.env.lambda` etc.) but that is messy too.


titiler/db/memcache.py
```diff
@@ -1,15 +1,27 @@
...
 from bmemcached import Client
 
+from titiler import settings
 from titiler.ressources.enums import ImageType
 
 
 class CacheLayer(object):
     """Memcache Wrapper."""
 
+    @classmethod
```
vincentsarago, line 13: ❤️

vincentsarago:
thanks @geospatial-jeff that's awesome. 

totally fine for the .env

PR #56, opened 2020-07-29
Add exception handlers

### LINKED ISSUE #41: Add exception handling
Currently the app either returns a 200 on successful request or server error (500) on unsuccessful request.  I think there is a lot of value in explicitly raising exceptions for error cases.  Most important in my opinion is `FileNotFound` and `TileNotFound` errors which both should return a `404` status code indicating that the particular file/tile was not found.  These status codes, if raised, may be used by the client to infer that there is no imagery available for the particular request and to fall back to another data source or provider.

I think the best way to do this is [global exception handlers](https://fastapi.tiangolo.com/tutorial/handling-errors/#install-custom-exception-handlers) which return a particular status code whenever a specific exception is raised.  This is inline with having [standardized exceptions](https://github.com/developmentseed/titiler/blob/master/titiler/errors.py) within the app (which is already implemented, although they aren't currently used anywhere).

geospatial-jeff:
- Adds fastapi exception handlers.

Closes #41

### DIFF
--- a/titiler/api/endpoints/stac.py
+++ b/titiler/api/endpoints/stac.py
@@ -18,15 +18,15 @@ from titiler.api.deps import (
     request_hash,
 )
 from titiler.db.memcache import CacheLayer
+from titiler.errors import BadRequestError
 from titiler.models.cog import cogBounds, cogInfo, cogMetadata
 from titiler.models.mapbox import TileJSON
 from titiler.ressources.enums import ImageMimeTypes, ImageType
 from titiler.ressources.responses import ImgResponse
 from titiler.templates.factory import web_template
 
-from fastapi import APIRouter, Depends, HTTPException, Path, Query
+from fastapi import APIRouter, Depends, Path, Query
 
-from starlette import status
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, Response
 
@@ -150,10 +150,7 @@ async def stac_tile(
     headers: Dict[str, str] = {}
 
     if not image_params.expression and not assets:
-        raise HTTPException(
-            status_code=status.HTTP_403_FORBIDDEN,
-            detail="Must pass Expression or Asset list.",
-        )
+        raise BadRequestError("Must pass Expression or Asset list.")
 
     tilesize = scale * 256
     tms = morecantile.tms.get(TileMatrixSetId.name)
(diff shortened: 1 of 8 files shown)


titiler/errors.py
```diff
@@ -11,3 +22,39 @@ class TileNotFoundError(TilerError):
 
 class BadRequestError(TilerError):
     """Bad request error."""
+
+
+DEFAULT_STATUS_CODES = {
```
geospatial-jeff, line 27: These are up for debate

titiler/api/endpoints/stac.py
```diff
@@ -150,10 +150,7 @@ async def stac_tile(
     headers: Dict[str, str] = {}
 
     if not image_params.expression and not assets:
-        raise HTTPException(
-            status_code=status.HTTP_403_FORBIDDEN,
-            detail="Must pass Expression or Asset list.",
-        )
+        raise BadRequestError("Must pass Expression or Asset list.")
```
vincentsarago, line 153: maybe we should let rio-tiler raise here https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/io/stac.py#L227 (this is in rio-tiler 2.0b2)

titiler/errors.py
```diff
@@ -1,4 +1,15 @@
 """Titiler error classes."""
+from typing import Callable, Dict, Type
+
+from cogeo_mosaic.errors import MosaicAuthError, MosaicError, MosaicNotFoundError
+from rasterio.errors import RasterioError, RasterioIOError
+from rio_tiler.errors import InvalidBandName, RioTilerError, TileOutsideBounds
```
vincentsarago, line 6: Maybe we can add basic rasterio errors like https://github.com/mapbox/rasterio/blob/master/rasterio/errors.py#L40

titiler/api/endpoints/stac.py
vincentsarago: @geospatial-jeff I fixed the exception issue in rio-tiler-crs, you can pin rio-tiler-crs~=3.0b2 https://pypi.org/project/rio-tiler-crs/3.0b2/

titiler/errors.py
vincentsarago: fine for now ;-)

titiler/api/endpoints/stac.py
```diff
@@ -446,10 +423,7 @@ async def stac_tilejson(
     kwargs.pop("maxzoom", None)
 
     if not expression and not assets:
-        raise HTTPException(
-            status_code=status.HTTP_403_FORBIDDEN,
-            detail="Expression or Assets HAVE to be set in the queryString.",
-        )
+        raise BadRequestError("Expression or Assets HAVE to be set in the queryString.")
```
vincentsarago, line 426: Should we raise `InvalidBandName` here as rio-tiler does?

vincentsarago:
It looks good, I think we can just make the change I proposed and it will be good ;-)

also I updated the changelog https://github.com/developmentseed/titiler/commit/d1ab1ae3174a0aeebf100039022881a0bf985517 make sure to pull the latest master and update a line about this pr 🙏

PR #57, opened 2020-07-29
Restructure app

### LINKED ISSUE #42: Remove intermediary api router
The fastapi router defined in `titiler.api.api` doesn't really add much.  I can see a use case for using an intermediary router if that router used a particular custom defined route class etc. but that isn't the case here.  It would be better to attach the routes directly to the fastapi app defined in `titiler.main`, remove `titiler.api.api`, and maybe rename `titiler.main` -> `titiler.app` to be more standard with fastapi conventions.

geospatial-jeff:
Closes #42


titiler/main.py
```diff
@@ -18,13 +20,36 @@
...
 )
+app.include_router(cog.router, prefix="/cog", tags=["Cloud Optimized GeoTIFF"])
+app.include_router(tms.router)
 add_exception_handlers(app, DEFAULT_STATUS_CODES)
+_include_extra_router(
+    app,
+    module="titiler.api.endpoints.stac",
+    prefix="/stac",
+    tags=["SpatioTemporal Asset Catalog"],
+)
```
vincentsarago, line 46: because of the new rio-tiler, stac is no more optional!

vincentsarago:
@geospatial-jeff what do you think about removing the entire `api/` directory?  

```
- /titiler/api/endpoints/* -> /titiler/endpoints/*
- /titiler/api/deps.py -> /titiler/dependencies.py
- /titiler/api/utils.py -> /titiler/utils.py
```

also what's your feeling on absolute vs relative import?

geospatial-jeff:
> @geospatial-jeff what do you think about removing the entire `api/` directory?

Big fan!  Thought about it when writing the PR but wanted the initial code to focus on #42.  

> also what's your feeling on absolute vs relative import?

Personally I like relative but I think absolute is more widely accepted as being the correct choice

vincentsarago:
> Big fan! Thought about it when writing the PR but wanted the initial code to focus on #42.

do you want to do it now or I can do it in another PR ?

PR #59, opened 2020-07-29
add generic jsonresponse

### LINKED ISSUE #43: add generic JSON response class
It looks like most routes which don't implement `ImgResponse` are manually setting cache-ttl response headers:

image

All of these routes (I think) are JSON responses, and should implement a similar `JSONResponse` object which automatically sets cache-control rather than defining within the scope of each request.  The `JSONResponse` object would be very similar to `ImgResponse`, but of course implementing a different protocol.

geospatial-jeff:
Closes #43

### DIFF
--- a/titiler/endpoints/cog.py
+++ b/titiler/endpoints/cog.py
@@ -27,7 +27,7 @@ from titiler.templates.factory import web_template
 from fastapi import APIRouter, Depends, Path, Query
 
 from starlette.requests import Request
-from starlette.responses import HTMLResponse, Response
+from starlette.responses import HTMLResponse, JSONResponse, Response
 from starlette.templating import Jinja2Templates
 
 router = APIRouter()
@@ -43,9 +43,8 @@ async def cog_bounds(
     resp: Response, url: str = Query(..., description="Cloud Optimized GeoTIFF URL."),
 ):
     """Return the bounds of the COG."""
-    resp.headers["Cache-Control"] = "max-age=3600"
     with COGReader(url) as cog:
-        return {"bounds": cog.bounds}
+        return JSONResponse({"bounds": cog.bounds})
 
 
 @router.get(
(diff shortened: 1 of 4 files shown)


titiler/endpoints/cog.py
```diff
@@ -27,7 +27,7 @@
 from fastapi import APIRouter, Depends, Path, Query
 
 from starlette.requests import Request
-from starlette.responses import HTMLResponse, Response
+from starlette.responses import HTMLResponse, JSONResponse, Response
```
vincentsarago, line 30: I think this `JSONResponse` should come from `titiler.ressources.responses`

titiler/endpoints/cog.py
```diff
@@ -405,6 +402,9 @@ async def cog_tilejson(
         if minzoom:
             center[-1] = minzoom
         tjson = {
+            "tilejson": "2.2.0",
+            "version": "1.0.0",
+            "scheme": "xyz",
```
vincentsarago, line 407: why do you need to add this, before it was taken care of by the model!

vincentsarago:
thanks @geospatial-jeff 

Quick question, what is restricting us to define the default response to our custom JSONResponse at the app level?

titiler/ressources/responses.py
```diff
@@ -13,21 +13,16 @@ class XMLResponse(Response):
...
-        """Image response."""
-        headers.update({"Content-Type": media_type})
-        if ttl:
-            headers.update({"Cache-Control": "max-age=3600"})
-        self.body = self.render(content)
-        self.status_code = 200
-        self.media_type = media_type
-        self.background = background
-        self.init_headers(headers)
+    def __init__(self, *args, ttl: int = 2600, **kwargs):
```
vincentsarago, line 16: I think `ttl` should be set to `3600` here!

vincentsarago:
Alright I gave this a though (and some search on internet)

IMO we should remove any custom thing from the endpoints (meaning we don't set custom response/header) but we do this at the application level using middleware

```python
@app.middleware("http")
async def header_middleware(request: Request, call_next):
    """Add custom header."""
    response = await call_next(request)
    response.headers["Cache-Control"] = "public,max-age=3600"
    return response
```

what do you think @geospatial-jeff

geospatial-jeff:
Closing in favor of #61

PR #61, opened 2020-07-30
add cache control middleware and remove useless Response

### LINKED ISSUE #43: add generic JSON response class

vincentsarago:
overtake #59 and closes #43 

cc @geospatial-jeff

### DIFF
--- a/tests/routes/test_cog.py
+++ b/tests/routes/test_cog.py
@@ -18,6 +18,7 @@ def test_bounds(reader, app):
     assert response.status_code == 200
     body = response.json()
     assert len(body["bounds"]) == 4
+    assert response.headers["Cache-Control"] == "public, max-age=3600"
 
 
 @patch("titiler.endpoints.cog.COGReader")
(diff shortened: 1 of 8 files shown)


tests/routes/test_cog.py
```diff
@@ -326,3 +329,5 @@ def test_tile_outside_bounds_error(reader, app):
     reader.side_effect = mock_reader
     response = app.get("/cog/tiles/15/0/0?url=https://myurl.com/cog.tif&rescale=0,1000")
     assert response.status_code == 404
+    # NOT THIS MIGHT CHANGE
+    assert response.headers["Cache-Control"] == "public, max-age=3600"
```
vincentsarago, line 333: Should we add `Cache-Control` to responses with 4** and 5** status code?

titiler/endpoints/cog.py
```diff
@@ -109,7 +101,7 @@ async def cog_metadata(
             "description": "Return an image.",
         }
     },
-    "response_class": ImgResponse,
+    "response_class": Response,
```
vincentsarago, line 104: There is no more need of the ImgResponse class (it was setting the cache control by default)

tests/routes/test_cog.py
geospatial-jeff: I think it makes sense to include for 4XX but not 5XX

kylebarron: 5XX is server error right? Definitely shouldn't cache those responses...

vincentsarago: cool, gonna skip cache for those

PR #63, opened 2020-07-31
add cog validation with rio-cogeo

### LINKED ISSUE #37: Add COG validation endpoint

geospatial-jeff:
Closes #37 

It would be sweet to update rio-cogeo to return warnings and errors (ref https://github.com/cogeotiff/rio-cogeo/issues/142)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,7 @@ inst_reqs = [
     "python-binary-memcached",
     "rio-color",
     "rio-tiler-crs==3.0b2",
+    "rio-cogeo",
 ]
 extra_reqs = {
     "dev": ["pytest", "pytest-cov", "pytest-asyncio", "pre-commit", "requests"],
--- a/titiler/endpoints/cog.py
+++ b/titiler/endpoints/cog.py
@@ -492,7 +492,7 @@ def validate_cog(
     strict: bool = Query(False, description="Treat warnings as errors"),
 ):
     """Validate a COG"""
-    return {"valid": cog_validate(url, strict=strict, quiet=True)}
+    return {"valid": cog_validate(url, strict=strict)}
 
 
 @router.get("/viewer", response_class=HTMLResponse, tags=["Webpage"])
--- a/titiler/endpoints/cog.py
+++ b/titiler/endpoints/cog.py
@@ -6,6 +6,7 @@ from typing import Any, Dict, Optional
 from urllib.parse import urlencode
 
 from rasterio.transform import from_bounds
+from rio_cogeo.cogeo import cog_validate
 from rio_tiler_crs import COGReader
 
 from titiler import utils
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -12,6 +12,7 @@
     "python-binary-memcached",
     "rio-color",
     "rio-tiler-crs==3.0b2",
+    "rio-cogeo",
```
kylebarron, line 15: I assume rio-cogeo isn't a large dependency?

titiler/endpoints/cog.py
```diff
@@ -485,6 +486,15 @@ def wmts(
     )
 
 
+@router.get("/validate", response_class=JSONResponse)
+def validate_cog(
+    url: str = Query(..., description="Cloud Optimized GeoTIFF URL."),
+    strict: bool = Query(False, description="Treat warnings as errors"),
+):
+    """Validate a COG"""
+    return {"valid": cog_validate(url, strict=strict)}
```
kylebarron, line 495: I think it would be helpful to return more output, such as whether windows exist. `rio-cogeo validate` allows for valid COG files to not have overviews, and just prints them as warnings.

setup.py
vincentsarago: no it's not

titiler/endpoints/cog.py
vincentsarago: yeah, I'd love a mix of `rio cogeo info` and `rio cogeo validate`. sadly you can't run both directly. maybe we should refactor rio_cogeo.cogeo.validate to return the `info` needed for `rio cogeo info`

kylebarron: Or `rio-cogeo info` to also call validate

vincentsarago: I'm not sure it's possible sadly. 
There is a `bug` in rasterio, once you call `rasterio.open('path', overview_level=1)` you can't open back the raw data. maybe this was fix 🤷‍♂️.

vincentsarago: well I just tested and it works 🤷‍♂️

geospatial-jeff: Once https://github.com/cogeotiff/rio-cogeo/pull/143/commits/7738517290fd3c0748b45b67594dfc2efa7c3168 and https://github.com/cogeotiff/rio-cogeo/pull/144 are released we should be good

titiler/endpoints/cog.py
```diff
@@ -6,6 +6,7 @@
 from urllib.parse import urlencode
 
 from rasterio.transform import from_bounds
+from rio_cogeo.cogeo import cog_validate
```
vincentsarago, line 9: @geospatial-jeff what do you think about using `cog_info` here?

geospatial-jeff: Makes sense to me with new rio-cogeo release

vincentsarago: the goal is to be able to print info in addition to `is_valid` (and not use /cog/info) 

I'll work on a small web page after this PR ;-)

vincentsarago:
done in #68

PR #66, opened 2020-07-31
update for latest cogeo-mosaic and rio-tiler

### LINKED ISSUE #64: Update for latest cogeo-mosaic / rio-tiler
right now there is a mismatch between rio-tiler version and it's leading the CI to fail!

### LINKED ISSUE #65: Replace Async code with latest MosaicBackend
For sake of code simplicity we will remove those to block 
https://github.com/developmentseed/titiler/blob/b6c8132727894356df7bbdc01620127554ccb08a/titiler/endpoints/mosaic.py#L198-L216

https://github.com/developmentseed/titiler/blob/b6c8132727894356df7bbdc01620127554ccb08a/titiler/endpoints/mosaic.py#L265-L301

The new MosaicBackends has a `.tile` and `.point` method that will take care of the data fetching directly. This is not async but we plan to make a AsyncMosaicBackends at one point

vincentsarago:
closes #64 
closes #65 

cc @geospatial-jeff

### DIFF
--- a/titiler/endpoints/mosaic.py
+++ b/titiler/endpoints/mosaic.py
@@ -1,17 +1,13 @@
 """API for MosaicJSON Dataset."""
 
-import asyncio
-import os
 import re
-from typing import AsyncGenerator, Coroutine, Dict, List, Optional, Tuple
+from typing import Dict, Optional
 from urllib.parse import urlencode
 
 import morecantile
-import numpy
 from cogeo_mosaic.backends import MosaicBackend
 from cogeo_mosaic.mosaic import MosaicJSON
 from cogeo_mosaic.utils import get_footprints
-from rio_tiler_crs import COGReader
 from rio_tiler_crs.cogeo import geotiff_options
 
 from titiler import utils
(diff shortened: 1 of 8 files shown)


titiler/errors.py
```diff
@@ -33,8 +43,10 @@ class BadRequestError(TilerError):
     TileOutsideBounds: status.HTTP_404_NOT_FOUND,
     TileNotFoundError: status.HTTP_404_NOT_FOUND,
     MosaicNotFoundError: status.HTTP_404_NOT_FOUND,
+    NoAssetFoundError: status.HTTP_404_NOT_FOUND,
     RasterioIOError: status.HTTP_404_NOT_FOUND,
-    InvalidBandName: status.HTTP_404_NOT_FOUND,
+    MissingAssets: status.HTTP_404_NOT_FOUND,
```
vincentsarago, line 48: should this return `status.HTTP_400_BAD_REQUEST` @geospatial-jeff ?

titiler/errors.py
```diff
@@ -33,8 +43,10 @@ class BadRequestError(TilerError):
...
+    InvalidAssetName: status.HTTP_404_NOT_FOUND,
```
vincentsarago, line 49: same ?

titiler/errors.py
geospatial-jeff: I think 400 is appropriate here since this exception will be raised if the caller doesn't provide assets/expression.

titiler/errors.py
geospatial-jeff: This could either be 400 or 404.  You can argue that it should be 404 because the asset doesn't exist in the STAC item, you can also argue that the caller should know what assets are present in the item so it should raise a 400.

Personally I'm leaning towards 400 but I don't think either is wrong.

vincentsarago: thanks! I'll keep 404 👌

titiler/endpoints/mosaic.py
```diff
@@ -189,34 +165,13 @@ async def mosaic_point(
 
     timings = []
     headers: Dict[str, str] = {}
+    threads = int(os.getenv("MOSAIC_CONCURRENCY", MAX_THREADS))
```
vincentsarago, line 168: @geospatial-jeff I've added back this ;-)

PR #71, opened 2020-08-06
[WIP] move business logic out of api routes

geospatial-jeff:
This PR moves business logic out of api routes.  This is the first step in transforming titiler from an application to a python library which builds an application.


vincentsarago:
we went another way in #79 ! Thanks @geospatial-jeff for the initial work 🙏

Issue #77, opened 2020-08-23
Use absolute paths for loading templates


geospatial-jeff:
Titiler fails to find location of html templates when installed as a python library because the current template paths are relative and not absolute (ex. https://github.com/developmentseed/titiler/blob/master/titiler/endpoints/stac.py#L35).  Using absolute paths to load html templates would fix this issue.

Issue #78, opened 2020-08-23
Use normal def to declare routes if no async code is called


geospatial-jeff:
I've run into an interesting race condition using titiler with arturo-stac-api.  For my use case, I am trying to mount titiler's stac and demo routers onto my stac application.  In the stac application I have an endpoint (`/collections/{collectionId}/items/{itemId}/tiles`) which, when called with a `text/html` accept header, sends a redirect request to the titiler's stac viewer.  The call chain looks like:

1. GET request to `/collections/{collectionId}/items/{itemId}/tiles`.
2. REDIRECT request to `/stac/viewer`
    2a. GET request to `/info`
    2b. GET request to `/tilejson.json`
    2c. GET request to `/tiles` (as tiles render in the browser).

**Each request in the browser (2a/2b/2c) does its own GET request to `/collections/{collectionId}/items/{itemId}` using the blocking `requests` library.

The race condition arises because the application is blocked during 2a/2b/2c as these requests require sending a GET request to the application to fetch the stac item which the application cannot process because it is currently blocked.

Switching all of the appropriate routes in titiler's stac router from `async def` to `def` fixed this issue, as the blocking call to `requests.get` to fetch the stac item is no longer blocking the main thread but instead run in an external threadpool.

geospatial-jeff:
Another solution would be to wrap blocking calls in starlette's `run_in_threadpool` but I think since the routes mostly contain blocking code it would be simpler to switch route declaration from `async def` to `def`

PR #79, opened 2020-08-24
refactor: create tiler factories

### LINKED ISSUE #33: reduce code duplication

vincentsarago:
closes #33 

should also solve #78 #77 

⚠️  This is a major refactor. ⚠️ 

This PR introduces 2 new Base Class `TilerFactory` and `MosaicTilerFactory`. Each one creates a router with expected endpoints (/tiles, /point, /tilejson ....). 

I've tried to add comments inside the class to help with the review.

```python
app = FastAPI()

cog = TilerFactory(router_prefix="cog")
app.include_router(cog.router, prefix="/cog", tags=["Cloud Optimized GeoTIFF"])
```

### ToDo
- [ ] write docs 

cc @geospatial-jeff


titiler/dependencies.py
```diff
@@ -44,201 +48,201 @@ async def request_hash(request: Request) -> str:
...
-class CommonTileParams:
-    """Common Tile parameters."""
+def TMSParams(
+    TileMatrixSetId: TileMatrixSetNames = Query(
+        TileMatrixSetNames.WebMercatorQuad,  # type: ignore
+        description="TileMatrixSet Name (default: 'WebMercatorQuad')",
+    )
+) -> morecantile.TileMatrixSet:
+    """TileMatrixSet Dependency."""
+    return morecantile.tms.get(TileMatrixSetId.name)
```
vincentsarago, line 58: This is a simple Callable dependency that returns a morecantil TMS model

titiler/dependencies.py
```diff
@@ -44,201 +48,201 @@ async def request_hash(request: Request) -> str:
...
-
-class CommonMetadataParams:
+def MosaicTMSParams(
+    TileMatrixSetId: MosaicTileMatrixSetNames = Query(
+        MosaicTileMatrixSetNames.WebMercatorQuad,  # type: ignore
+        description="TileMatrixSet Name (default: 'WebMercatorQuad')",
+    )
+) -> morecantile.TileMatrixSet:
+    """TileMatrixSet Dependency."""
+    return morecantile.tms.get(TileMatrixSetId.name)
```
vincentsarago, line 68: Mosaic only allow WebMercator for now!

titiler/dependencies.py
```diff
@@ -44,201 +48,201 @@ async def request_hash(request: Request) -> str:
...
+
+
+@dataclass
+class DefaultDependency:
+    """Dependency Base Class"""
+
+    kwargs: dict = field(init=False, default_factory=dict)
```
vincentsarago, line 75: The DefaultDependency class is more a placeholder than a real dependency for now!

titiler/dependencies.py
```diff
@@ -44,201 +48,201 @@ async def request_hash(request: Request) -> str:
...
+
+
+@dataclass
```
vincentsarago, line 78: we now use dataclass for dependency definition

titiler/dependencies.py
```diff
@@ -44,201 +48,201 @@ async def request_hash(request: Request) -> str:
...
+        if parsed.scheme.split("+")[-1] == "mosaicid":
+            self.url = f"{settings.DEFAULT_MOSAIC_BACKEND}{settings.DEFAULT_MOSAIC_HOST}/{parsed.netloc}.json.gz"
+
+        # if parsed.scheme.startswith("landsat+"):
+        #     self.reader = L8Reader
+        #     self.url = self.url.replace("landsat+", "")
+        # elif parsed.scheme.startswith("sentinel2cog+"):
+        #     self.reader = Sentinel2COG
+        #     self.url = self.url.replace("sentinel2cog+", "")
+        # ....
```
vincentsarago, line 100: The PathParams is maybe the most important dependency. I've added a `reader` placeholder to add more fancy things like `landsat+mosaicid`

titiler/dependencies.py
```diff
@@ -44,201 +48,201 @@ async def request_hash(request: Request) -> str:
...
+        None,
+        title="Asset indexes",
+        description="comma (',') delimited asset names (might not be an available options of some readers)",
+    )
+    kwargs: dict = field(init=False, default_factory=dict)
+
+    def __post_init__(self):
+        """Post Init."""
+        if self.assets is not None:
+            self.kwargs["assets"] = self.assets.split(",")
```
vincentsarago, line 117: ⚠️ we use kwargs to store the `assets` option so we can just pass `**params.kwargs` to the function (in the tiler) and we do not have to worry if the reader accept the asset keyword. 

in the tiler factory with use `add_asset_deps` to specify if we use `AssetsParams` or `DefaultDependency`. The deps will populate the open api with assets (if set to true) but on the function we just need to declare 

```python
        optionsParams = AssetsParams if self.add_asset_deps else DefaultDependency
        
        def info(
            src_path=Depends(self.path_dependency), 
            options: optionsParams = Depends()
        ):
            """Return basic info."""
            with reader(src_path.url) as src_dst:
                info = src_dst.info(**options.kwargs)

```

titiler/endpoints/demo.py
```diff
@@ -1,15 +1,16 @@
...
 
 from starlette.requests import Request
 from starlette.responses import HTMLResponse
 from starlette.templating import Jinja2Templates
 
 router = APIRouter()
-
-
-templates = Jinja2Templates(directory="titiler/templates")
+template_dir = pkg_resources.resource_filename("titiler", "templates")
```
vincentsarago, line 12: fixes #77

titiler/main.py
```diff
@@ -18,26 +24,40 @@
...
+
+
+@cog.router.get("/validate", response_model=RioCogeoInfo)
+def cog_validate(
+    src_path: PathParams = Depends(),
+    strict: bool = Query(False, description="Treat warnings as errors"),
+):
+    """Validate a COG"""
+    return rio_cogeo_info(src_path.url, strict=strict)
+
```
vincentsarago, line 44: The above is equivalend of extending the TilerFactory Class

```python
from dataclasses import dataclass

@dataclass
class CustomTiler(TilerFactory):
    router_prefix = "cog"
    def __post_init__(self):
        super().__post_init__()
        self._add_validation()
    def _add_validation(self):
        @self.router.get(
            "/validate",
            response_model=RioCogeoInfo,
            name=f"{self.router_prefix}info",
        )
        def validate(
            src_path: PathParams = Depends(),
            strict: bool = Query(False, description="Treat warnings as errors"),
        ):
            """Validate a COG"""
            return rio_cogeo_info(src_path.url, strict=strict)

cog = CustomTiler()
app.include_router(cog.router, prefix="/cog", tags=["Cloud Optimized GeoTIFF"])
```

titiler/dependencies.py
geospatial-jeff: If its not meant to be instantiated maybe make it an `ABC`

titiler/dependencies.py
```diff
@@ -44,201 +48,201 @@ async def request_hash(request: Request) -> str:
...
+
+@dataclass
+class DefaultDependency:
+    """Dependency Base Class"""
+
+    kwargs: dict = field(init=False, default_factory=dict)
+
+
+@dataclass
+class PathParams(DefaultDependency):
```
geospatial-jeff, line 79: Should this be called `QueryParams` instead?  I don't see any path parameters defined.

titiler/dependencies.py
```diff
@@ -44,201 +48,201 @@ async def request_hash(request: Request) -> str:
...
+    """Create dataset path from args"""
+
+    url: str = Query(..., description="Dataset URL")
+    reader: Optional[Type[BaseReader]] = field(init=False, default=None)  # Placeholder
+
+    def __post_init__(self,):
```
geospatial-jeff, line 85: Using `__post_init__` here is serving the same purpose as defining a dependency (ex. [`TMSParams`](https://github.com/developmentseed/titiler/pull/79/files#diff-7e64e756c81e1d4282046f8561c647ebR51-R55)) just with a different pattern.  Would it be possible to use either one or the other everywhere you need to modify a request parameter after it is declared?

vincentsarago: Yeah except if we go ahead with `reader` defined as a scheme (e.g `landsat+...`) It's not really used right now but the Factory expects the PathParams class to have 2 properties (url+reader).

titiler/dependencies.py
vincentsarago: @geospatial-jeff I'm not sure to understand

geospatial-jeff: The class is called `PathParams` but it only defines query parameters below (`url: str = Query(...)`).  So this class doesn't seem to be related to path parameters at all.

geospatial-jeff: Oh I see, its called `PathParams` because it creates the dataset path, and I'm confusing that with http path params

PR #81, opened 2020-08-24
switch to github actions

vincentsarago:


vincentsarago:
CI is failing because of change in rio-tiler (fixed in #79)

PR #85, opened 2020-08-26
make memcached an optional dependency

vincentsarago:
This PR does:
- rename `titiler.db` submodule to `titiler.cache`
- make `memcached` an optional requirements
- move `get_cache` from `titiler.utils` to `titiler.dependencies`
- simplify cache method (get/set) to enable more cache options (not just imagery)


vincentsarago:
Nope let's get rid of caching! 
Will add an example how to implement caching at app level later

PR #106, opened 2020-09-09
add env option to baseFactory to set GDAL environment variable

vincentsarago:
sometimes you need to work with dataset that needs specific environment variables to be set (e.g Landsat, data in requester-pays bucket ...)

### DIFF
--- a/titiler/endpoints/factory.py
+++ b/titiler/endpoints/factory.py
@@ -7,6 +7,7 @@ from typing import Any, Callable, Dict, Optional, Type, Union
 from urllib.parse import urlencode
 
 import pkg_resources
+import rasterio
 from cogeo_mosaic.backends import BaseBackend, MosaicBackend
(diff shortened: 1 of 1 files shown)


vincentsarago:
Note: This solution will only work if the reader doesn't use threads

https://gist.github.com/vincentsarago/82aac58f4a0d32238b9c885bec216b6f

Somehow rasterio/gdal get lost in multiple layers of threads (non async endpoints runs in threads https://github.com/encode/starlette/blob/master/starlette/concurrency.py#L21). I think this is fine for now (will need documentation).

The only solution I see is to add `env` option directly in multithreaded readers (https://github.com/cogeotiff/rio-tiler/blob/master/rio_tiler/io/base.py#L96 and https://github.com/cogeotiff/rio-tiler-pds/blob/master/rio_tiler_pds/reader.py#L33), same for mosaic 🤷‍♂️

PR #112, opened 2020-09-24
Valid stack name in documentation; including example for requester pays

fredliporace:
Minor documentation updates suggestions based on my deploy experience. The STACK_NAME example contains an invalid character. I'm also including an example on how to define extra ENV vars for lambda, I needed to include that to work with CBERS' requester pays buckets.


vincentsarago (approved):

vincentsarago:
👼 thanks @fredliporace!

PR #116, opened 2020-09-25
Brotli compression middleware

kylebarron:
Use brotli compression by default, unless `br` is not in the `Accept-Encoding` headers, in which case fall back to gzip.

In a future PR it would be good to remove compression middleware from WebP, PNG, JPEG, and TIF responses. (So compression is used for NPY, JSON, and HTML).


vincentsarago:
@kylebarron would you mind adding tests please!

titiler/middleware.py
```diff
@@ -43,3 +48,70 @@ async def dispatch(self, request: Request, call_next):
...
+                    self.lgblock,
+                    self.minimum_size,
+                )
+                await responder(scope, receive, send)
+                return
+
+            if "gzip" in headers.get("Accept-Encoding", ""):
+                responder = GZipResponder(self.app, self.minimum_size)
+                await responder(scope, receive, send)
+                return
```
vincentsarago, line 115: Not required, but would be nice to add `zlib` and `deflate` https://github.com/vincentsarago/lambda-proxy/blob/master/lambda_proxy/proxy.py#L602-L611

titiler/middleware.py
```diff
@@ -43,3 +48,70 @@ async def dispatch(self, request: Request, call_next):
...
+            Range is 16 to 24. If set to 0, the value will be set based on the
+            quality.
+        minimum_size: Only compress responses that are bigger than this value in bytes.
+        """
+        self.app = app
+        self.quality = quality
+        self.mode = getattr(BrotliMode, mode)
+        self.minimum_size = minimum_size
+        self.lgwin = lgwin
+        self.lgblock = lgblock
```
vincentsarago, line 93: I don't know what to think about the options that are brotli specific... maybe we can use `kwargs` here and let the user forward what ever options they want to the Responder!

titiler/middleware.py
kylebarron: From a little reading it seemed like gzip was always preferable to zlib and deflate. The algorithms are very similar, so there's not much difference in compression size

vincentsarago:
> In a future PR it would be good to remove compression middleware from WebP, PNG, JPEG, and TIF responses. (So compression is used for NPY, JSON, and HTML).

I know @geospatial-jeff proposed a solution over slack, but I tihnk a simple solution here would be to add a `set` of `exclude` response type for which we don't want to compress again.

https://github.com/developmentseed/titiler/blob/master/titiler/ressources/common.py#L15-L19

titiler/middleware.py
kylebarron: We could use `kwargs` for most things except Mode. `BrotliResponder` needs `mode` to be an instance of `Brotli.Mode`. Or we can just make the mode always generic.

vincentsarago: I'm fine with generic mode, if a user wants something custom he/she can create another middleware

kylebarron:
> I tihnk a simple solution here would be to add a `set` of `exclude` response type for which we don't want to compress again.

I agree. That's pretty much what I was looking for, but didn't know how to implement it

vincentsarago:
> I agree. That's pretty much what I was looking for, but didn't know how to implement it

I think the response is not available while in the `CompressMiddleware` but it is inside the `responder`https://github.com/encode/starlette/blob/41218ac546a3c7a0f1982f2677177ce9a0ab3616/starlette/middleware/gzip.py#L23 

anyway, let's take care of this later

kylebarron:
@vincentsarago I rolled back the `kwargs` change because the `BrotliResponder` has [no default params](https://github.com/fullonic/brotli-asgi/blob/708e57ae4d0cd5101361de500dc642119d2d1890/brotli_asgi/__init__.py#L76-L84). So you can't pass `kwargs`

vincentsarago:
>  I rolled back the kwargs change because the BrotliResponder has no default params. So you can't pass kwargs

what about something like:

```python
defaul_brotli_config = {
    "quality": 4,
    "mode": "MODE_TEXT",
    "lgwin": 22,
    "lgblock": = 0,
    "minimum_size": = 400,
}

class CompressMiddleware:
    """Brotli middleware public interface."""

    def __init__(
        self, app: ASGIApp, minimum_size: int = 400, brotli_config: Dict = None) -> None:
    ...
    self.brotli_config = {**defaul_brotli_config, **brotli_config} if brotli_config else defaul_brotli_config

    ...

    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
        responder = BrotliResponder(
            self.app,
            self.brotli_config["quality"],
            self.brotli_config["mode"],
            self.brotli_config["lgwin"],
            self.brotli_config["lgblock"],
            self.minimum_size,
        )
        ...
```

kylebarron:
🎉

PR #120, opened 2020-09-29
update ECS config and notes

### LINKED ISSUE #119: Expensive NAT charges when deployed to ECS
I deployed titiler to ECS, never touched it (besides making sure it worked), and then four days later checked and had $150 in NAT charges, with **3TB** of "Data Processed by NAT Gateways" at $0.045/GB. The egress bandwidth is only 1.694 GB, however, so nearly all of this data was inbound to the tiler.

The logs all look like this:
image

There are such logs nearly every second:
image

It looks similar to this issue: https://github.com/tiangolo/fastapi/issues/253. There are several configuration settings in the fastapi [docker options](https://hub.docker.com/r/tiangolo/uvicorn-gunicorn-fastapi).

vincentsarago:
closes #119

### DIFF
--- a/docs/deployment/ecs.md
+++ b/docs/deployment/ecs.md
@@ -1,6 +1,7 @@
 # AWS ECS (Fargate) + ALB (Application Load Balancer)
 
-**⚠** When using Fargate or vanilla ECS, you should set the number of worker carefully (https://github.com/developmentseed/titiler/issues/119 and https://github.com/tiangolo/fastapi/issues/253).
+!!! warning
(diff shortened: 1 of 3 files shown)


vincentsarago:
@kylebarron @geospatial-jeff could you review this please!

docs/deployment/ecs.md
```diff
@@ -1,5 +1,7 @@
 # AWS ECS (Fargate) + ALB (Application Load Balancer)
 
+**⚠** When using Fargate or vanilla ECS, you should set the number of worker carefully (https://github.com/developmentseed/titiler/issues/119 and https://github.com/tiangolo/fastapi/issues/253).
```
kylebarron, line 3: This would be perfect for an "admonition"

```
!!! warning
	When using Fargate or vanilla ECS...
```

image

I'd also include something like "you should set # of workers carefully. Setting too high a number of workers could lead to extra charges due to a bug in fastapi"

kylebarron (approved):

PR #122, opened 2020-10-01
revert to gzip compression only

### LINKED ISSUE #121: Revert Brotli compression support
Sadly Brotli is not yet supported by Mangum (see: https://github.com/jordaneremieff/mangum/pull/140)

Also @kylebarron directly added Gzip Fallback to brotli-asgi (`==1.0.0`) so we don't need our custom CompressMiddleware

vincentsarago:
closes #121


kylebarron:
I think it would've been fine to use mangum master for now until there's a new release

PR #142, opened 2020-11-06
Adapt for change in rio-tiler, cogeo-mosaic and morecantile

vincentsarago:
status: waiting for rio-tiler 2.0.0rc and cogeo-mosaic==3.0.0a17 to be published


setup.py
```diff
@@ -38,7 +37,7 @@
 
 setup(
     name="titiler",
-    version="0.1.0a8",
+    version="0.1.0a9",
```
vincentsarago, line 40: 🤦 I reset my local branch one commit too early.
FYI: I'm going going to change this, I think it's fine!

PR #154, opened 2020-11-23
[WIP] add AsyncReader support

vincentsarago:
This needs discussion! This class is mostly to help the integration of https://github.com/geospatial-jeff/aiocogeo-tiler

ref: https://github.com/developmentseed/titiler/issues/78

### DIFF
--- a/tests/test_asynccog.py
+++ b/tests/test_asynccog.py
@@ -1,0 +1,404 @@
+"""test AsyncTilerFactory."""
+
+from io import BytesIO
+from unittest.mock import patch
+
+import numpy
+import pytest
+from rio_tiler.io import COGReader
+from rio_viz.compat import AsyncReader
+
+from titiler.endpoints import factory
+
+from .conftest import mock_rasterio_open, parse_img
+
+from fastapi import FastAPI
+
+from starlette.testclient import TestClient
+
+AsyncCOGReader = type("AsyncReader", (AsyncReader,), {"reader": COGReader})
+router = factory.TilerFactory(reader=COGReader, router_prefix="cog").router
+
+
+@pytest.fixture(autouse=True)
+def async_app(set_env) -> TestClient:
+    """Create App."""
+    app = FastAPI()
+    app.include_router(router, prefix="/cog")
+    return TestClient(app)
+
+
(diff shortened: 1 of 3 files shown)


titiler/endpoints/factory.py
```diff
@@ -1081,3 +1081,580 @@ def point(
             )
 
             return {"coordinates": [lon, lat], "values": values}
+
+
+# ref: https://github.com/python/mypy/issues/5374
+@dataclass  # type: ignore
+class AsyncBaseFactory(metaclass=abc.ABCMeta):
```
vincentsarago, line 1088: I'm not sure there is a better solution than copying the whole code from `non-async` classes. Def not great from a maintainer POV!

kylebarron: Why not use mixins?

vincentsarago: because it doesn't solve the property definition order.  The problem here, is there is no `AsyncReader` (like COGReader) available by default so we need to set `reader: Type[AsyncBaseReader]` without default. In rio-tiler we overcome this kind of problem using attrs ... so I guess we could do the same here

tests/test_asynccog.py
```diff
@@ -0,0 +1,404 @@
...
+
+from titiler.endpoints import factory
+
+from .conftest import mock_rasterio_open, parse_img
+
+from fastapi import FastAPI
+
+from starlette.testclient import TestClient
+
+AsyncCOGReader = type("AsyncReader", (AsyncReader,), {"reader": COGReader})
```
vincentsarago, line 19: @geospatial-jeff not sure if you'll have time, but while doing your aiocogeo (async) vs rio-tiler (sync) benchmark it would be nice to also benchmark async rio-tiler using `rio-viz` `AsyncReader` compat class.

```python
from rio_tiler.io import COGReader
from rio_viz.compat import AsyncReader

AsyncCOGReader = type("AsyncReader", (AsyncReader,), {"reader": COGReader})

async with AsyncCOGReader("my.tif") as cog:
     img = await cog.tile(....)
```

vincentsarago:
let's revisit this later

PR #155, opened 2020-11-25
Minor documentation updates

fredliporace:
A couple of minor things I found while playing with custom CMAPs and TMS.

I'm not 100% sure about the custom TMS example fix, I'm using morecantile 1.3.1 here but checked against the latest source and is seems to apply to that version also.


docs/concepts/customization.md
```diff
@@ -235,7 +235,7 @@ EPSG6933 = morecantile.TileMatrixSet.custom(
 )
 
 # 2. Register TMS
-morecantile.tms.register(custom_tms.EPSG6933)
+morecantile.tms.register(custom_tms=EPSG6933)
```
vincentsarago, line 238: this changed recently. To register a custom TMS you need to do
```python
from morecantile import tms

tms = tms.register(EPSG6933)
```

https://github.com/developmentseed/titiler/blob/09448211e95f406526701a31064a659460e4bbb7/titiler/dependencies.py#L24

titiler/custom/cmap.py
```diff
@@ -1,7 +1,7 @@
 """
 Titiler Custom colormaps.
 
-Custom colormaps needs to be register in titiler/api/deps.py
+Custom colormaps needs to be register in titiler/dependencies.py
 e.g `cmap.register("above", custom_colormap.above_cmap)`
```
vincentsarago, line 5: we can change the commend with
```
e.g cmap = cmap.register({"above": custom_colormap.above_cmap})
```

docs/concepts/customization.md
vincentsarago: Note: this changed in morecantile 2.0

titiler/custom/cmap.py
vincentsarago: this changed in rio-tiler==2.0.0rc1 I think

fredliporace:
Hi @vincentsarago I updated to take into account the new versions but was not able to test the custom TMS changes - I'm still working with titiler 0.1.0a12... I may revisit this when I update to the newer version next week, but maybe it is good to go as it is.

vincentsarago:
thanks @fredliporace that's perfect 🙏

Issue #166, opened 2020-12-09
add `/info.geojson` to return a GeoJSON feature from a dataset


vincentsarago:
The endpoint should return a geojson feature for a dataset

PR #171, opened 2020-12-09
WIP: sketch geojson support

vincentsarago:
ref: #166

### DIFF
--- a/titiler/endpoints/factory.py
+++ b/titiler/endpoints/factory.py
@@ -1,6 +1,7 @@
 """TiTiler Router factories."""
 
 import abc
+import json
 import os
 from dataclasses import dataclass, field
 from typing import Any, Callable, Dict, Optional, Type, Union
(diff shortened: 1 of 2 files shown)


titiler/endpoints/factory.py
```diff
@@ -181,19 +181,83 @@ def info(self):
...
             "/info",
-            response_model=Info,
+            response_model=Union[Info, Dict],
             response_model_exclude={"minzoom", "maxzoom", "center"},
             response_model_exclude_none=True,
             responses={200: {"description": "Return dataset's basic info."}},
         )
         def info(
             src_path=Depends(self.path_dependency),
+            accept: Optional[JsonType] = Header(None),
```
vincentsarago, line 197: tests are failing because by default we send `content="*/*"` which is not a member of `JsonType`

Having a defined accept values plays nice with OpenAPI but is too restrictive (?).

I'm not sure what's the best here, IMO just having `as_geojson=True` option would be simpler 🤷 

cc @geospatial-jeff

geospatial-jeff: hmm I've never been a fan of how fastapi handles `*/*`.  Also you can't subclass enums which makes things difficult.

I think maybe separate `info.json` and `info.geojson` routes makes the most sense for now.  It's definitely less standard but we already are using a pattern that doesn't rely on the accept header but instead the postfix of the route (ex `/tilejson.json`).  So I'd say let's keep going with that pattern to keep things consistent.

I would definitely prefer an accept header longer term (its more standard HTTP), but maybe that is outside the scope of this PR.  Switching to the use of an accept header to dictate api behavior seems like a larger topic.

vincentsarago: 🙏 thanks @geospatial-jeff `.json` `.geosjon` 👌

Issue #11, opened 2020-04-04
add EKS ? 


vincentsarago:
explore deployment using EKS https://aws.amazon.com/fr/eks/

geospatial-jeff:
can this be closed with #178?

PR #177, opened 2020-12-13
[WIP] helm chart

geospatial-jeff:
Ref #11 

Adds a helm chart for deploying titiler to k8s.  Right now the chart is very stripped down and doesn't do much besides pull the docker container and configuring container resources/env vars.

You can try this out as so (requires `minikube`, `helm`, and `kubectl`):
```bash
minikube start
kubectl config use-context minikube
helm init --wait

# in the k8s directory
helm install -f titiler/Chart.yaml titiler
```


vincentsarago:
finished this in #182 🙏 @geospatial-jeff

Issue #192, opened 2021-01-27
move `validate` in MosaicTilerFactory class


vincentsarago:
https://github.com/developmentseed/titiler/blob/39108ba40502351c7b03720b0dcd874806d78ad4/titiler/endpoints/mosaic.py#L10-L13

vincentsarago:
closed in #206

PR #206, opened 2021-02-02
Move 'validate' in MosaicTilerFactory class

drnextgis:
Fix for #192


vincentsarago:
Thanks for this pr @drnextgis 🙏

PR #209, opened 2021-02-03
Switch examples to folium

### LINKED ISSUE #135: switch demo to folium
sadly ipyleaflet doesn't work in binder nor in colab 👎

### LINKED ISSUE #208: Create_CustomSentinel2Tiler.ipynb notebook is broken
* in the first section "Create the CustomTiler" instead of `from titiler.models.dataset import Info, Metadata` should be `from rio_tiler.models import Info, Metadata`
* `optquery` variable is not defined

drnextgis:
Fixes https://github.com/developmentseed/titiler/issues/135
closes #135 
closes #208 

Folium doesn't support custom CRS, [so this](https://github.com/developmentseed/titiler/blob/master/docs/examples/Working_with_nonWebMercatorTMS.ipynb) notebook hasn't been changed. Also I've removed all outputs from notebooks to make them consistent.


drnextgis:
Would someone review it please?

vincentsarago:
@drnextgis yep, I'm doing this right now, there are couple of change to do

vincentsarago:
@drnextgis Thanks for this PR it's really appreciated 

> Would someone review it please?

As I answered, I was doing it but it took a bit more time because I had to fix issue with the code, which then lead me to ask if you had run the notebook yourself? Not a big issue but just took me a bit more time. Reviewing Notebook is not something easy because diff is not enough. 


I'm trying to review PR/issue as they comes but I'm also managing other project so I hope you'll understand if there is some lag between PR and review ;-). As an open-source project we value everyone contribution and time.

vincentsarago (approved):

PR #213, opened 2021-02-04
Bump rio-tiler dep to 2.0.0

kylebarron:


setup.py
```diff
@@ -11,7 +11,7 @@
     "python-dotenv",
     "morecantile",
     "rio-cogeo>=2.1,<2.2",
-    "rio-tiler>=2.0.0rc4,<2.1",
+    "rio-tiler>=2.0.0,<2.1",
```
vincentsarago, line 14: let's do 2.0.1 for the minimal version https://github.com/cogeotiff/rio-tiler/blob/master/CHANGES.md#201-2021-02-04

PR #214, opened 2021-02-07
Use local CDK instance in docs

kylebarron:
The current deployment docs instruct to install CDK globally (for both Python, using pip, and Node, using NPM). This isn't a great practice in general, but especially so for CDK. If the versions for CDK's Node and Python bindings don't match up exactly, CDK will raise an exception and exit. This PR helps enforce best practices by isolating the Node CDK installation. It creates a `package.json` file with `cdk` as a dependency, then aliases `cdk` in the `scripts` field, which means that `npm run cdk` or `yarn cdk` will find the local CDK instance, not the global one. 

This PR doesn't suggest anything for isolating the CDK python packages since there are many ways to achieve that: Conda/Virtualenvs/Pipenv/Poetry.


vincentsarago:
thanks @kylebarron, indeed this is really nice

deployment/aws/package-lock.json
```diff
@@ -0,0 +1,1436 @@
+{
```
vincentsarago, line 1: @kylebarron I don't think we have to add the package-lock here 🤷‍♂️. having package-lock will mean that the contributor has to run npm install before pushing to the repo (in case of change of cdk dependency). I think it's a step we can skip ?

kylebarron: I think it's fine to remove. Hopefully AWS never makes a broken cdk release 😄

PR #218, opened 2021-02-10
remove default create/update endpoint in MosaicTilerFactory

### LINKED ISSUE #217: Remove Mosaic Write operations
By default we add / (POST) and / (PUT) to the mosaicjson endpoint to either create or update a mosaicjson document. The problem is that if you add this to a AWS deployment and do not set your IAM correctly, someone could create mosaic on you S3 bucket.

I don't think the default implementation is good and I think we should remove them (and add examples)

https://github.com/developmentseed/titiler/blob/d329f2c0eb0b385db065ff890706792ec14a79d0/titiler/endpoints/factory.py#L743-L794

vincentsarago:
closes #217

### DIFF
--- a/docs/concepts/mosaics.md
+++ b/docs/concepts/mosaics.md
@@ -9,7 +9,6 @@ Ref: https://github.com/developmentseed/mosaicjson-spec
 
 [TODO]
 
-
 ### Links
 
 - https://medium.com/devseed/cog-talk-part-2-mosaics-bbbf474e66df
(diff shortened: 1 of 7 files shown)


docs/concepts/mosaics.md
```diff
@@ -9,6 +9,7 @@ Ref: https://github.com/developmentseed/mosaicjson-spec
 
 [TODO]
 
+
```
vincentsarago, line 12:
```suggestion
```

PR #232, opened 2021-02-12
make timings and metadata headers optional

### LINKED ISSUE #229: Tiler `timings` should be optional!
https://github.com/developmentseed/titiler/blob/f2821b9421ecad0e1a79356fc82c3f6ee39833f5/titiler/endpoints/factory.py#L339-L341

vincentsarago:
closes #229

### DIFF
--- a/titiler/endpoints/factory.py
+++ b/titiler/endpoints/factory.py
@@ -98,6 +98,9 @@ class BaseTilerFactory(metaclass=abc.ABCMeta):
     # Add specific GDAL environement (e.g {"AWS_REQUEST_PAYER": "requester"})
     gdal_config: Dict = field(default_factory=dict)
 
+    # add info (e.g timings) to the response headers
+    debug: bool = field(default=False)
+
     def __post_init__(self):
         """Post Init: register route and configure specific options."""
         self.register_routes()
@@ -336,9 +339,10 @@ class TilerFactory(BaseTilerFactory):
                 )
             timings.append(("format", round(t.elapsed * 1000, 2)))
 
-            headers["Server-Timing"] = ", ".join(
-                [f"{name};dur={time}" for (name, time) in timings]
-            )
+            if self.debug:
+                headers["Server-Timing"] = ", ".join(
+                    [f"{name};dur={time}" for (name, time) in timings]
+                )
 
             return Response(content, media_type=format.mimetype, headers=headers)
 
@@ -539,9 +543,10 @@ class TilerFactory(BaseTilerFactory):
                         )
             timings.append(("dataread", round(t.elapsed * 1000, 2)))
 
-            response.headers["Server-Timing"] = ", ".join(
-                [f"{name};dur={time}" for (name, time) in timings]
-            )
+            if self.debug:
+                response.headers["Server-Timing"] = ", ".join(
+                    [f"{name};dur={time}" for (name, time) in timings]
+                )
 
             return {"coordinates": [lon, lat], "values": values}
 
@@ -601,7 +606,7 @@ class TilerFactory(BaseTilerFactory):
                 )
             timings.append(("format", round(t.elapsed * 1000, 2)))
 
-            if timings:
+            if self.debug:
                 headers["Server-Timing"] = ", ".join(
                     [f"{name};dur={time}" for (name, time) in timings]
                 )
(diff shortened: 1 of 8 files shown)


titiler/endpoints/factory.py
```diff
@@ -878,11 +883,11 @@ def tile(
...
-            headers["Server-Timing"] = ", ".join(
-                [f"{name};dur={time}" for (name, time) in timings]
-            )
-
-            headers["X-Assets"] = ",".join(data.assets)
+            if self.debug:
+                headers["Server-Timing"] = ", ".join(
+                    [f"{name};dur={time}" for (name, time) in timings]
+                )
+                headers["X-Assets"] = ",".join(data.assets)
```
kylebarron, line 890: I wouldn't necessarily agree that `X-Assets` should be considered a debugging header. `Server-Timing` would never be used on the frontend, but `X-Assets` could actually be potentially used in the frontend to provide information about what images went into the tile.

kylebarron (approved):
I was just thinking the other day that these should be optional! For example the [MDN article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing#privacy_and_security) says

> The `Server-Timing` header may expose potentially sensitive application and infrastructure information. Consider to control which metrics are returned when and to whom on the server side. For example, you could only show metrics to authenticated users and nothing to the public.

vincentsarago:
@kylebarron I've added `add_assets_in_headers` option in the MosaicTilerFactory. Do it make senses to rename `debug` -> `add_timings_in_headers` then ?

kylebarron:
You could just have an `optional_headers` argument that takes a list. Then it'll be more extensible for the future

vincentsarago:
@kylebarron I'm not sure to fully understand what you mean. 

are you saying something like 👇 
```python
cog = TilerFactory(optional_headers={"timings", ...}
```

kylebarron:
Yeah basically. You could have an enum of available optional headers. It might also make sense to make the keys the same as the header names, e.g.
```py
cog = TilerFactory(optional_headers={"Server-Timing", "X-Assets", ...}
```
(though should be case insensitive)

vincentsarago:
> You could have an enum of available optional headers. 

well in this case we will need 2 different enums (x-assets is only available for mosaic). I feel this will add more code than we want. 

I'll commit some change, let me know after it what you think

vincentsarago:
in fact I'm going to take care of #234 before finishing this

vincentsarago:
Alright @kylebarron, 

> You could have an enum of available optional headers.

I don't see how this will be easy to do. basically right now we have `"Server-Timing"` for basic tilers and `"Server-Timing", "X-Assets"` for Mosaic tilers. First I'm not sure how to implement this using enums and second we will need two different `enums` (one only with `"Server-Timing"` and another with `"Server-Timing" and "X-Assets"`).

I wanted to use `attrs` so we could have used the attribute `validation` but attrs makes things a bit more complex (https://github.com/developmentseed/titiler/issues/234)

I don't really want to spend more time on it but feel free to propose your own PR 🙏 


Note: 
for now I've went the way of 2 different attributes:
- `add_timing_headers`
- `add_assets_headers`

kylebarron:
Not every factory needs to implement all the available optional headers... You can have a single enum listing the optional headers available, and just note in the docs (or maybe give a warning at runtime) if a non-mosaic tiler attempts to use the `X-Assets` header.

vincentsarago:
> You can have a single enum listing the optional headers available, and just note in the docs (or maybe give a warning at runtime) if a non-mosaic tiler attempts to use the X-Assets header.

seems like more code to me! but also no idea how enum will work for this case when we have multiple values provided as input

kylebarron:
Well the enum part isn't important... All I was suggesting is something like this, instead of making a separate parameter for every optional header like `add_timing_headers`, `add_assets_headers`, ...

```py
tilerfactory(
	...,
	optional_headers=['Server-Timing', 'X-Assets'])
```

```py
from . import OptionalHeaders
tilerfactory(
	...,
	optional_headers=[OptionalHeaders.Server_Timing, OptionalHeaders.X_Assets])
```

vincentsarago:
> All I was suggesting is something like this, instead of making a separate parameter for every optional header like add_timing_headers, add_assets_headers, ...

I ear you, and what I'm saying is just that IMO this will require more code/tests ... so I'm just trying to find the best solution here!

vincentsarago:
@kylebarron could you provide a PR with your solution ?

vincentsarago:
Alright it took me some time but I think I understand what you proposed! can you check the last commit and 👍 👎 @kylebarron

kylebarron (approved):

Issue #240, opened 2021-02-18
Docs page with explanation of relevant GDAL environment variables


kylebarron:
#### Problem description

There are many GDAL environment variables, and it's hard to know the implications of each of them just from reading the GDAL documentation.

I think I can start a PR on this. Possible start:

- `CPL_VSIL_CURL_ALLOWED_EXTENSIONS`: A list of extensions that GDAL is allowed to open. For example if set to `.tif`, then GDAL wouldn't open any other files. For example, it wouldn't open JPEG2000 files with a `.jp2` extension, but also wouldn't open Geotiffs exposed through an API endpoint that don't have a `.tif` suffix.
- `GDAL_HTTP_MULTIPLEX`. Only applies with HTTP2, which not all servers support, but when set to `YES` and uses HTTP2, can download multiple ranges in parallel. 
- `GDAL_INGESTED_BYTES_AT_OPEN`: Gives the number of bytes GDAL should read for COG headers. 
	- Explain how to see header of average dataset with `aiocogeo info`. Vincent says you can't get it from GDAL.
- `GDAL_DISABLE_READDIR_ON_OPEN`: Depends on the source data.
- `GDAL_HTTP_MERGE_CONSECUTIVE_RANGES`: Set to `YES` to merge ranges when possible. 

Unknown env vars:

- CPL_VSIL_CURL_NON_CACHED
- VSI_CACHE

Ref: https://trac.osgeo.org/gdal/wiki/ConfigOptions

vincentsarago:
`CPL_VSIL_CURL_NON_CACHED` this one is to make sure nothing is cached for a path or a pattern. I use it here https://github.com/developmentseed/tilebench/blob/master/tilebench/viz.py#L75-L78

> Starting with GDAL 2.3, the CPL_VSIL_CURL_NON_CACHED configuration option can be set to values like /vsicurl/http://example.com/foo.tif:/vsicurl/http://example.com/some_directory, so that at file handle closing, all cached content related to the mentioned file(s) is no longer cached. This can help when dealing with resources that can be modified during execution of GDAL related code. Alternatively, VSICurlClearCache() can be used.


`CPL_VSIL_CURL_CHUNK_SIZE` can also be a nice one

[...]

PR #241, opened 2021-02-19
Start of GDAL config settings doc

kylebarron:
Closes https://github.com/developmentseed/titiler/issues/240

I'd love help for: GDAL_CACHEMAX, GDAL_DATA, PROJ_LIB.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -28,6 +28,7 @@ nav:
       - Tiler Factories: "advanced/tiler_factories.md"
       - Dependencies: "advanced/dependencies.md"
       - Customization: "advanced/customization.md"
+      - Performance Tuning: "advanced/performance_tuning.md"
       - APIRoute and environment variables: "advanced/APIRoute_and_environment_variables.md"
 
   - Default Endpoints:
--- a/docs/advanced/performance_tuning.md
+++ b/docs/advanced/performance_tuning.md
@@ -1,0 +1,177 @@
+## Overview
+
+Titiler makes use of several great underlying libraries, including [GDAL][gdal]
+and [Python bindings to GDAL][rasterio]. An effective deployment of titiler
+generally requires tweaking GDAL configuration settings. This document provides
+an overview of relevant settings. Full documentation from GDAL is available
+[here][gdal_config_docs].
+
+[gdal]: https://gdal.org/
+[rasterio]: https://rasterio.readthedocs.io/
+[gdal_config_docs]: https://gdal.org/user/configoptions.html
+
+## GDAL Configuration
+
+### Setting a config variable
+
+GDAL configuration is modified using environment variables. Thus in order to
+change a setting you'll need to set environment variables through your
+deployment mechanism. For example, in order to test locally you'd set an
+environment variable in bash:
+
+```bash
+export GDAL_HTTP_MULTIPLEX=YES
+```
+
+### Available configuration settings
+
+#### `GDAL_HTTP_MERGE_CONSECUTIVE_RANGES`
+
+When set to `YES`, this tells GDAL to merge adjacent range requests. Instead of
+making two requests for byte ranges `1-5` and `6-10`, it would make a single
+request for `1-10`. This should always be set to `YES`.
+
+#### `GDAL_DISABLE_READDIR_ON_OPEN`
+
+This is a very important setting to control the number of requests GDAL makes.
+
+This setting has two options: `FALSE` and `EMPTY_DIR`. `FALSE` (the default)
+causes GDAL to try to establish a list of all the available files in the
+directory. `EMPTY_DIR` tells GDAL to imagine that the directory is empty except
+for the requested file.
+
+When reading datasets with necessary external sidecar files, it's imperative to
+set `FALSE`. For example, the `landsat-pds` bucket on AWS S3 contains GeoTIFF
+images where overviews are in external `.ovr` files. If set to `EMPTY_DIR`, GDAL
+won't find the `.ovr` files.
+
+However, in all other cases, it's much better to set `EMPTY_DIR` because this
+prevents GDAL from making a `LIST` request.
+
+This setting also has cost implications for reading data from requester-pays
+buckets. When set to `FALSE`, GDAL makes a `LIST` request every time it opens a
+file. Since `LIST` requests are much more expensive than `GET` requests, this
+can bring unexpected costs.
+
+#### `CPL_VSIL_CURL_ALLOWED_EXTENSIONS`
+
+A list of file extensions that GDAL is allowed to open. For example if set to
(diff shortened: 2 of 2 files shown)


docs/deployment/gdal_config.md
```diff
@@ -0,0 +1,129 @@
...
+
+Titiler makes use of several great underlying libraries, including [GDAL][gdal]
+and [Python bindings to GDAL][rasterio]. An effective deployment of titiler
+generally requires tweaking GDAL configuration settings. This document provides
+an overview of relevant settings. Full documentation from GDAL is available
+[here][gdal_config_docs].
+
+[gdal]: https://gdal.org/
+[rasterio]: https://rasterio.readthedocs.io/
+[gdal_config_docs]: https://trac.osgeo.org/gdal/wiki/ConfigOptions
```
vincentsarago, line 11: https://gdal.org/user/configoptions.html

docs/deployment/gdal_config.md
```diff
@@ -0,0 +1,129 @@
...
+deployment mechanism. For example, in order to test locally you'd set an
+environment variable in bash:
+
+```bash
+export GDAL_HTTP_MULTIPLEX=YES
+```
+
+## Available configuration settings
+
+#### `GDAL_HTTP_MULTIPLEX`
```
vincentsarago, line 26: Maybe we could order the section in a kind of `priority`. I usually set HTTP_MULTIPLEX but I
 m 99.9% that it's useless because no cloud provider storage supports http2 for now. 

IMO the order could be :
- `GDAL_HTTP_MERGE_CONSECUTIVE_RANGES`
- `GDAL_DISABLE_READDIR_ON_OPEN`
- `CPL_VSIL_CURL_ALLOWED_EXTENSIONS`
- `GDAL_INGESTED_BYTES_AT_OPEN`
- `GDAL_CACHEMAX`
- `VSI_CACHE`
- `AWS_REQUEST_PAYER`
- Other cloud provider env
- PROJ_NETWORK (only for goal 3.2 / proj 7.1)

kylebarron: Yeah I was def going to ask what you thought the order should be. I just kinda threw ideas down randomly. Also maybe would be good to rename the doc to "config" with a GDAL section, and then a separate section for AWS with `AWS_REQUEST_PAYER`.

mkdocs.yml
```diff
@@ -54,6 +54,7 @@ nav:
       - enums: api/titiler/resources/enums.md
 
   - Deployment:
+    - GDAL Configuration: "deployment/gdal_config.md"
```
vincentsarago, line 57: Maybe this could go in the Advanced User Guide

docs/deployment/gdal_config.md
vincentsarago: 👍

docs/deployment/gdal_config.md
```diff
@@ -0,0 +1,129 @@
...
+          PROFILE
+            ...
+            Header size:      32770
+```
+
+It's wise to inspect the header sizes of your data sources, and set
+`GDAL_INGESTED_BYTES_AT_OPEN` appropriately. Beware, however, that the given
+number of bytes will be read for every image, so you don't want to make the
+value too large.
+
```
vincentsarago, line 87: might be nice to plug tilebench and https://github.com/developmentseed/tilebench#starlette-middleware

vincentsarago:
🥳 thanks so much @kylebarron, this is really cool 🙏

docs/deployment/gdal_config.md
```diff
@@ -0,0 +1,129 @@
...
+#### `GDAL_DATA`
+
+#### `PROJ_LIB`
+
+#### `VSI_CACHE`
+
+Setting this to `TRUE` enables GDAL to use an internal caching mechanism. It's
+strongly recommended to set this to `TRUE`.
+
+#### `VSI_CACHE_SIZE`
```
geospatial-jeff, line 127: should mention that this is a per-file handle.  Ex. if you open a VRT with 10 files and your VSI_CACHE_SIZE is 10 bytes, the total cache memory usage would be 100 bytes.

docs/deployment/gdal_config.md
```diff
@@ -0,0 +1,129 @@
...
+file. Since `LIST` requests are much more expensive than `GET` requests, this
+can bring unexpected costs.
+
+#### `GDAL_HTTP_MERGE_CONSECUTIVE_RANGES`
+
+When set to `YES`, this tells GDAL to merge adjacent range requests. Instead of
+making two requests for byte ranges `1-5` and `6-10`, it would make a single
+request for `1-10`. This should always be set to `YES`.
+
+#### `GDAL_CACHEMAX`
```
geospatial-jeff, line 116: disabling this causes issues when accessing the same block at the same time.  Also before GDAL2.4, when specified as a percentage, the size of the cache was computed based on the memory of the host, potentially causing memory leaks when used in a container

https://github.com/OSGeo/gdal/issues/640

geospatial-jeff:
Maybe we need a short paragraph explaining the high level strategy for making titiler fast, which is minimizing number of HTTP GET range requests.  And these GDAL environment variables can be tuned to your use case to do that.

docs/deployment/gdal_config.md
```diff
@@ -0,0 +1,129 @@
...
+
+However, in all other cases, it's much better to set `EMPTY_DIR` because this
+prevents GDAL from making a `LIST` request.
+
+This setting also has cost implications for reading data from requester-pays
+buckets. When set to `FALSE`, GDAL makes a `LIST` request every time it opens a
+file. Since `LIST` requests are much more expensive than `GET` requests, this
+can bring unexpected costs.
+
+#### `GDAL_HTTP_MERGE_CONSECUTIVE_RANGES`
```
geospatial-jeff, line 110: I think its important to explain why this matters - reduces number of HTTP requests, bytes transferred is less important than number of requests etc.

geospatial-jeff:
One AWS specific gotcha I've run into in the past is related to how GDAL looks up aws creds.  If they aren't stored in (1) the environment or (2) `~/.aws/credentials`, GDAL will check the EC2 instance profile (if applicable) which can be quite slow - adds up to ~1 second of latency per network call.

geospatial-jeff:
I bet it's also worth mentioning`CPL_CURL_VERBOSE` + `CPL_DEBUG` as a good debug tool (also tilebench).

docs/deployment/gdal_config.md
RichardScottOZ: Yes, definitely.  Might need more than a short paragraph.

vincentsarago:
👋 I've updated some part. I'm going to merge it now but anyone is free to continue updating the doc ;-)

PR #246, opened 2021-02-23
Issue#244

BnJam:
Include numpy import
Fix plt.imshow(...) where plt prefix was missing.


vincentsarago:
@BnJam thanks for the PR. 

I think most of the Import should be covered by `%pylab inline`

I think the main problem here is we don't ask for matplotlib to be installed

BnJam:
oh, you're right! Feel free to close/delete this issue/pr @vincentsarago

PR #247, opened 2021-02-24
add backend options to the MosaicTilerFactory

vincentsarago:
In cogeo-mosaic backend we are using `backend_options` to pass addition backend options (e.g client) 

cc @kylebarron

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,20 +3,20 @@ repos:
     rev: 19.10b0
     hooks:
       - id: black
-        language_version: python3.7
+        language_version: python3.8
         args: ["--safe"]
 
   - repo: https://github.com/PyCQA/isort
     rev: 5.4.2
     hooks:
       - id: isort
-        language_version: python3.7
+        language_version: python3.8
 
   - repo: https://github.com/PyCQA/flake8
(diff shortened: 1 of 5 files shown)


titiler/endpoints/factory.py
```diff
@@ -1101,6 +1111,7 @@ def tile(
                         src_path.url,
                         reader=self.dataset_reader,
                         reader_options=self.reader_options,
```
kylebarron, line 1113: Do you want to remove passing the `reader_options` to the backend?

vincentsarago: no we still need this because https://github.com/developmentseed/cogeo-mosaic/blob/master/cogeo_mosaic/backends/base.py#L156

kylebarron (approved):

.pre-commit-config.yaml
```diff
@@ -3,20 +3,20 @@ repos:
     rev: 19.10b0
     hooks:
       - id: black
-        language_version: python3.7
+        language_version: python3.8
```
kylebarron, line 6: 🎉 New M1 new Python version? 😉

vincentsarago: 🙈

PR #258, opened 2021-03-04
use singular for enums

vincentsarago:
cc @kylebarron

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@
 
 **breaking change**
 
-* renamed `OptionalHeader`, `MimeTypes` and `ImageDrivers` to the singular form.
+* renamed `OptionalHeaders`, `MimeTypes` and `ImageDrivers` to the singular form.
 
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -6,6 +6,10 @@
 * update FastAPI requirements
 * update minimal python version to 3.6
 
+**breaking change**
+
+* renamed `OptionalHeader`, `MimeTypes` and `ImageDrivers` to the singular form.
```
kylebarron, line 11:
```suggestion
* renamed `OptionalHeaders`, `MimeTypes` and `ImageDrivers` to the singular form.
```

kylebarron: Also nit: the term "MIME Type" is deprecated in favor of "Media Type" I believe

kylebarron (approved):

CHANGES.md
vincentsarago: let's go all in

PR #259, opened 2021-03-05
add `**render_params.kwargs` ito pass custom render params

vincentsarago:
this PR add `**render_params.kwargs` to pass custom render params in image.render method.

ref: https://github.com/developmentseed/titiler/discussions/255

cc @neda-dtu

### DIFF
--- a/tests/test_CustomRender.py
+++ b/tests/test_CustomRender.py
@@ -1,0 +1,63 @@
+# """Test TiTiler Custom Render Params."""
+
+from dataclasses import dataclass
+from typing import Optional, Union
+
+import numpy
+
+from titiler.dependencies import RenderParams
+from titiler.endpoints import factory
+
+from .conftest import DATA_DIR, parse_img
+
+from fastapi import FastAPI, Query
+
+from starlette.testclient import TestClient
+
+
+@dataclass
+class CustomRenderParams(RenderParams):
+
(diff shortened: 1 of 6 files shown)


tests/test_CustomRender.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def __post_init__(self):
+        super().__post_init__()
+        if self.output_nodata is not None:
+            self.kwargs["nodata"] = (
+                numpy.nan if self.output_nodata == "nan" else float(self.output_nodata)
+            )
+        if self.output_compression is not None:
+            self.kwargs["compress"] = self.output_compression
+
```
vincentsarago, line 36: cc @neda-dtu 
this is an example of custom RenderParams that add `output_nodata` and `output_compression` to the available query options (you could also hardcode some options).

titiler/dependencies.py
```diff
@@ -72,7 +72,7 @@ def TMSParams(
 class DefaultDependency:
     """Dependency Base Class"""
 
-    kwargs: dict = field(init=False, default_factory=dict)
+    kwargs: Dict = field(init=False, default_factory=dict)
```
vincentsarago, line 75: fix

geospatial-jeff (approved):

PR #262, opened 2021-03-05
add Custom Colormap querystring option

### LINKED ISSUE #153: User specified "custom" colormap
Is there a way to allow the client to specify a custom colormap "dynamically"?  For example, by providing a series of pixel+color pairs and dynamically generating a linear gradient colormap on the fly during rendering?

If such a capability doesn't exist, would you have any advice on how it might be best implemented?

vincentsarago:
closes #153  

This PR does:
- rename `color_map` query parameter to `colormap_name` 
- add `colormap` option.

The `colormap` option accepts JSON encoded object in form of:

```
{
    "1": {RGB(A)}
}
```

the `{RBG(A)}` value can either be a list of R,G,B (and Alpha) values or an hex color representation

cc @kylebarron @blazetopher

### DIFF
--- a/titiler/dependencies.py
+++ b/titiler/dependencies.py
@@ -10,7 +10,7 @@ import numpy
 from morecantile import tms
 from morecantile.models import TileMatrixSet
 from rasterio.enums import Resampling
-from rio_tiler.colormap import cmap
+from rio_tiler.colormap import cmap, parse_color
 from rio_tiler.errors import MissingAssets, MissingBands
 
 from .custom import cmap as custom_colormap
@@ -67,53 +67,6 @@ def TMSParams(
 ) -> TileMatrixSet:
     """TileMatrixSet Dependency."""
     return tms.get(TileMatrixSetId.name)
-
-
-def parse_color(value: Union[List[int], str]) -> List[int]:
-    """Parse RGB/RGBA color and return valid rio-tiler compatible RGBA colormap entry.
-
-    Args:
-        value (str or list of int): HEX encoded or list RGB or RGBA colors.
-
-    Returns:
-        list: RGBA values.
-
-    Examples:
-        >>> parse_color("#FFF")
-        [255, 255, 255, 255]
-
-        >>> parse_color("#FF0000FF")
-        [255, 0, 0, 255]
-
-        >>> parse_color("#FF0000")
-        [255, 0, 0, 255]
-
-        >>> parse_color([255, 255, 255])
-        [255, 255, 255, 255]
-
-    """
-    if isinstance(value, str):
-        hex_pattern = (
-            r"^#"
-            r"(?P<red>[a-fA-F0-9]{1,2})"
-            r"(?P<green>[a-fA-F0-9]{1,2})"
-            r"(?P<blue>[a-fA-F0-9]{1,2})"
-            r"(?P<alpha>[a-fA-F0-9]{1,2})?"
-            r"$"
-        )
-        match = re.match(hex_pattern, value)
-        if match:
-            factor = 1 if len(match.groups()[0]) == 2 else 2
-            value = [
-                int(n * factor, 16) for n in match.groupdict().values() if n is not None
-            ]
-        else:
-            raise Exception(f"Invalid color {value}")
-
-    if len(value) == 3:
-        value += [255]
-
-    return value
 
 
 def ColorMapParams(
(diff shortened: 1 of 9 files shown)


titiler/dependencies.py
```diff
@@ -68,12 +69,67 @@ def TMSParams(
     return tms.get(TileMatrixSetId.name)
 
 
+def parse_color(value: Union[List[int], str]) -> List[int]:
```
vincentsarago, line 72: I think I'm going to move this to rio-tiler

titiler/dependencies.py
```diff
@@ -68,12 +69,67 @@ def TMSParams(
...
+    if len(value) == 3:
+        value += [255]
+
+    return value
+
+
 def ColorMapParams(
-    color_map: ColorMapName = Query(None, description="Colormap name",)
+    colormap_name: ColorMapName = Query(None, description="Colormap name"),
+    colormap: str = Query(None, description="JSON encoded custom Colormap"),
```
vincentsarago, line 121: **breaking change**

titiler/dependencies.py
```diff
@@ -68,12 +69,67 @@ def TMSParams(
...
 ) -> Optional[Dict]:
     """Colormap Dependency."""
-    if color_map:
-        return cmap.get(color_map.value)
+    if colormap_name:
+        return cmap.get(colormap_name.value)
+
+    if colormap:
+        return json.loads(
```
vincentsarago, line 128: should we also accept base64 encoded string?

titiler/dependencies.py
vincentsarago: https://github.com/cogeotiff/rio-tiler/pull/361

titiler/dependencies.py
geospatial-jeff: I think base64 is less weird than JSON for get requests

vincentsarago: another way maybe would be to pass a list of value like 
```
colormap=1,#FFFF00&colormap=2,#FF00FF

#equivalent of

colormap={"1": "#FFFF00", "2":"#FF00FF"}
```

vincentsarago:
following some discussion on slack, 

Right now a user could define a **discrete** colormap by passing a JSON encoded object in form of 
```
colormap={"1": "#FFFF00", "2":"#FF00FF"}
``` 

I understand that this is quite uncommon for GET request to receive `dict` in a query parameter, but we don't really have a choice (because we can't store a colormap `somewhere`. 

we need a way to pass a `list` of key/value, and for now this PR uses a JSON dict.

`List` seems to be an accepted type for query parameter, using the form `opt=1&opt=2` (FastAPI will resolve this to `opt=[1, 2]`. So maybe a form like `colormap=1,#FFFF00&colormap=2,#FF00FF` would be better 🤷 


## Options 

- JSON encoded object {note: would need to be decoded and re-encoded in tilejson endpoint)

`...preview.png?colormap={"1": "#FFFF00", "2":"#FF00FF"}`  (should be url encoded)

- base64 encoded dict (seems to be a nice option but pretty custom)

`...preview.png?colormap=dsaidhauhdasiojhd928e1uqwijs09qus0x8ahxas`

- List defined colormap value (restricted to hex color definition and custom key/color `1,#fff` form)

`...preview.png?colormap=1,#FFFF00&colormap=2,#FF00FF`

👂

vincentsarago:
Note: the `#` shouldn't be used in a non-encoded url

vincentsarago:
#### Option 1
JSON encoded object {note: would need to be decoded and re-encoded in tilejson endpoint)

     ...preview.png?colormap={"1": "#FFFF00", "2":"#FF00FF"} (should be url encoded)

#### Option 2
base64 encoded dict (seems to be a nice option but pretty custom)
    
    ...preview.png?colormap=dsaidhauhdasiojhd928e1uqwijs09qus0x8ahxas

#### Option 3
List defined colormap value (restricted to hex color definition and custom key/color 1,#fff form)

     ...preview.png?colormap=1,FFFF00&colormap=2,FF00FF

I think it's fine to go with **option 1**. It's not the best solution but:
- Starlette will take care of the urlencode/decode part, so we don't really need to worry in the API
- using base64 encoded dict in query string is not something widely used/accepted
- The third option looks more like proper HTTP query string but still custom because we will need to encode `2,FF00FF`

Notes:
I also though about something like 
``` 
colormap=FFFF00&colormap_stop=1,&colormap= FF00FF&colormap_stop=2
``` 

which will 💯 match the specs but still really verbose and not as `easy` as option 1

cc @kylebarron @geospatial-jeff @ingalls

PR #284, opened 2021-03-30
create titiler namespaces packages 

vincentsarago:
ref: https://github.com/developmentseed/titiler/discussions/278

### Why
As mentioned in #278, TiTiler is much more than an application, it's a python module that lets user create custom Tiler. 

I believe that splitting `core` and `application` could help users to have a better understanding but also help us to move quickly on some features. 

### How

- **titiler.core**: core functions (e.g tiler factories)

Tiler factories, ressources and models

- **titiler.application**: `demo` application

basic application (adds Uvicorn, python-dotenv and brotli dependencies) 

## main changes 
- split core functionalities and demo application
- move `custom` (colormap, tms) to the demo application
- removed unused function `titiler.utils.get_hash`
- switch to `pkg_resources` to locate templates files

##### other
- use a bash script instead of tox to do PyPI publishing 

## To Do
- talk 
- update docs
- update readmes


.pre-commit-config.yaml
```diff
@@ -37,8 +37,8 @@ repos:
           ]
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.770
+    rev: v0.812
     hooks:
       - id: mypy
         language_version: python3.8
-        args: ["--no-strict-optional", "--ignore-missing-imports"]
+        args: ["--no-strict-optional", "--ignore-missing-imports", "--explicit-package-bases", "--namespace-packages"]
```
vincentsarago, line 44: ref: https://github.com/pre-commit/mirrors-mypy/issues/40#issuecomment-785000979

titiler/application/setup.py
```diff
@@ -0,0 +1,45 @@
...
+    long_description = f.read()
+
+inst_reqs = [
+    "titiler-core",
+    "brotli-asgi>=1.0.0",
+    "python-dotenv",
+]
+extra_reqs = {
+    "test": ["pytest", "pytest-cov", "pytest-asyncio", "requests"],
+    "server": ["uvicorn[standard]>=0.12.0,<0.14.0"],
```
vincentsarago, line 15: Uvicorn is optional (e.g when running the application in AWS Lambda you don't need Uvicorn)

titiler/core/titiler/core/utils.py
```diff
@@ -1,18 +1,11 @@
...
-import hashlib
-import json
 import time
-from typing import Any, Dict, Optional, Tuple
+from typing import Dict, Optional, Tuple
 
 from geojson_pydantic.features import Feature
 
 
-def get_hash(**kwargs: Any) -> str:
```
vincentsarago, line 11: this function wasn't used in titiler

tox.ini
```diff
@@ -2,9 +2,18 @@
...
-    python -m pytest --cov titiler --cov-report xml --cov-report term-missing --ignore=venv
+deps =
+    pytest
+    pytest-cov
+    pytest-asyncio
+    requests
+skip_install = true
+commands_pre =
+    python -m pip install titiler/core
+    python -m pip install titiler/application
```
vincentsarago, line 13: because titiler-application depends on titiler-core, we install them manually

tox.ini
```diff
@@ -2,9 +2,18 @@
...
+commands =
+    python -m pytest titiler/core --cov titiler.core --cov-report xml --cov-report term-missing --ignore=venv
+    python -m pytest titiler/application --cov titiler.application --cov-report xml --cov-report term-missing --ignore=venv
```
vincentsarago, line 16: TODO: cov reports get overwritten

vincentsarago:
Still awaiting for feedback but if we move forward with namespaces here is IMO what we could do:
- move mosaic features into `titiler.mosaic`
- add `rio-tiler-mvt` feature (https://github.com/developmentseed/titiler-mvt) into `titiler.raster_to_vector`

### Maybe
- `titiler.mvt` instead of https://github.com/developmentseed/timvt
- `titiler.public_datasets` instead of https://github.com/developmentseed/titiler-pds
- `titiler.stac_api` (WIP) a mosaic backend based on stac_fastapi


cc @geospatial-jeff @kylebarron

.pre-commit-config.yaml
kylebarron: Yes this is important

kylebarron:
Quick notes from chat

- Decide whether to use `-` or `.` in PyPI names, i.e. `pip install titiler.core` or `pip install titiler-core`. (AWS CDK uses `.`, Google uses `-`, I use `.`)
- Should follow up on whether you can use `mypy` in pre-commit still after moving to namespaces. In my experience, any linters/code formatters that are purely static are a good fit for pre-commit (black, isort, yapf, pylint, etc). If the tool needs to follow imports (mypy, pytest), then it's simplest to run in each module individually.
- Ideally put all config in config files, e.g. `mypy.ini`, `setup.cfg`, `pyproject.toml` (my preferred, since it's a PEP standard). Then whether run inside or outside pre-commit should give the same output.

tox.ini
kylebarron: Use
```
coverage run --parallel-mode
```
in each module then
```
coverage combine modules/*
```
I don't know how it works through `pytest-cov`; I use `coverage` directly.

kylebarron (approved):

scripts/publish
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
kylebarron, line 1: Nit: `#! /usr/bin/env bash` is most portable https://stackoverflow.com/q/16365130

setup.cfg
```diff
@@ -3,17 +3,53 @@ current_version = 0.2.0
 commit = True
 tag = True
 tag_name = {new_version}
-parse = 
+parse =
 	(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)
```
kylebarron, line 7: You use spaces everywhere else; is tabs intentional here?

setup.cfg
```diff
@@ -3,17 +3,53 @@ current_version = 0.2.0
...
 	{major}.{minor}.{patch}{pre}{prenum}
 	{major}.{minor}.{patch}
 
-[bumpversion:file:setup.py]
+[bumpversion:titiler/application:setup.py]
 search = version="{current_version}"
 replace = version="{new_version}"
 
-[bumpversion:file:titiler/__init__.py]
+[bumpversion:file:titiler/application/titiler/application/version.py]
```
kylebarron, line 17: Intentional to switch between `bumpversion` and `bumpversion:file`?

titiler/application/tests/routes/test_mosaic.py
```diff
@@ -18,7 +18,7 @@
 def mosaic_read_factory(fname: str) -> Callable:
     """Factory method for patching mosaic reading"""
 
-    def _read() -> MosaicJSON:
+    def _read(*args, **kwargs) -> MosaicJSON:  # type: ignore
```
kylebarron, line 21: I just do `*args: Any, **kwargs: Any` I think Pydantic does that most of the time too
https://github.com/samuelcolvin/pydantic/blob/3f849a368f260caab3288f2ad8fe018d7e903bc3/pydantic/decorator.py#L50-L52

.github/workflows/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         if: success()
         uses: codecov/codecov-action@v1
         with:
-          file: ./coverage.xml
+          file: ./coverage_core.xml,./coverage_mosaic.xml,./coverage_application.xml,
```
kylebarron: I haven't used codecov before, but this might give you three different reports on the dashboard, instead of a single report with three different modules.

PR #293, opened 2021-04-17
Minor text edits

abarciauskas-bgse:
Just made some minor text edits and added a question about the link

### DIFF
--- a/docs/advanced/APIRoute_and_environment_variables.md
+++ b/docs/advanced/APIRoute_and_environment_variables.md
@@ -6,8 +6,8 @@
     cog = TilerFactory(reader=COGReader, router_prefix="cog", gdal_config={"GDAL_DISABLE_READDIR_ON_OPEN":"FALSE"})
     ```
 
-Sometime, specifically when using GDAL, it can be usefull to have environment variables set for some endpoints
-(e.g when using Landsat data on AWS you need `GDAL_DISABLE_READDIR_ON_OPEN=FALSE` but you don't want this for your other endpoints). To be able to do this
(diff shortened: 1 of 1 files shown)


docs/advanced/APIRoute_and_environment_variables.md
```diff
@@ -6,8 +6,8 @@
     cog = TilerFactory(reader=COGReader, router_prefix="cog", gdal_config={"GDAL_DISABLE_READDIR_ON_OPEN":"FALSE"})
     ```
 
-Sometime, specifically when using GDAL, it can be usefull to have environment variables set for some endpoints
-(e.g when using Landsat data on AWS you need `GDAL_DISABLE_READDIR_ON_OPEN=FALSE` but you don't want this for your other endpoints). To be able to do this
+Sometimes, specifically when using GDAL, it can be useful to have environment variables set for certain endpoints
+(e.g. when using Landsat data on AWS you need `GDAL_DISABLE_READDIR_ON_OPEN=FALSE` but you don't want this environment variable set for other endpoints). To be able to do this
 we created a *custom* APIRoute class which wraps classic fastapi APIRoute with a `rasterio.Env()` block: https://github.com/developmentseed/titiler/blob/8a7127ca56631c2c327713d99e80285048c3aa6c/titiler/custom/routing.py#L13-L41
```
abarciauskas-bgse, line 11: should this link to https://github.com/developmentseed/titiler/blob/master/titiler/custom/routing.py#L14-L47 instead of the specific commit?

vincentsarago: the problem linking to master is that if the file gets updated we will loose the lines!

PR #300, opened 2021-04-21
Return 500 for `RasterioIOError`

kylebarron:


vincentsarago:
@kylebarron would you be able to update the tests 🙏 and also to give a bit more info why this change is needed ;-)

kylebarron (commented):
Looks good

PR #303, opened 2021-04-25
Patch example notebooks

abarciauskas-bgse:


docs/examples/notebooks/Working_with_STAC_simple.ipynb
```diff
@@ -129,6 +129,7 @@
     "    f\"{titiler_endpoint}/stac/info\",\n",
     "    params = {\n",
     "        \"url\": stac_item,\n",
+    "        \"assets\": \"pan\"\n",
```
abarciauskas-bgse, line 132: I believe `assets` is a required parameter, looking at the `stac/info` docs

vincentsarago: Yes it is now, but the version deployed in api.cogeo.xyz is not up to date 😬

PR #315, opened 2021-05-12
Helm - Add support for serviceAccountName

AndreaGiardini:
Hey

Small PR to add support for setting a serviceAccountName in the helm chart

### DIFF
--- a/deployment/k8s/titiler/values.yaml
+++ b/deployment/k8s/titiler/values.yaml
@@ -47,6 +47,8 @@ resources:
     cpu: 1
     memory: 2Gi
 
+serviceAccountName: ""
+
 nodeSelector: {}
 
 tolerations: []
--- a/deployment/k8s/titiler/templates/deployment.yaml
+++ b/deployment/k8s/titiler/templates/deployment.yaml
@@ -42,6 +42,9 @@ spec:
               port: http
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+      {{- with .Values.serviceAccountName }}
+      serviceAccountName: {{ . | quote }}
+      {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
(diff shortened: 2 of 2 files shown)


deployment/k8s/titiler/values.yaml
```diff
@@ -47,6 +47,8 @@ resources:
     cpu: 1
     memory: 2Gi
 
+serviceAccountName: ""
```
vincentsarago, line 50: should we put a default here? what happens if it's empty `""` ? 

I'm not helm expert 🤷‍♂️

AndreaGiardini: Hey @vincentsarago 

If it is an empty string then it is not rendered:
```
titiler/deployment/k8s/titiler on  master [⇡?]
➜ helm template . | grep serviceAcc

titiler/deployment/k8s/titiler on  master [⇡?]
✘1 ➜ vim values.yaml
```
If it has a value (e.g. test):
```
titiler/deployment/k8s/titiler on  master [⇡!?]
➜ helm template . | grep serviceAcc
      serviceAccountName: "test"
titiler/deployment/k8s/titiler on  master [⇡?]
```

vincentsarago: would it be ok to add just a small comment explaining why this variable is useful? 

I mean I don't use k8s so I don't really understand what the `serviceAccountName` is used for 😄

AndreaGiardini: I could do that if you wish but I believe any k8s engineer will know what serviceaccounts are and why they are needed. Most of the helm charts that you see out there have this variable.

In a few words, they are used to limit what a certain pod can / can't do. In my case, I use a serviceaccount to give `titiler` read-only access to our GCS bucket rather than read-write.

vincentsarago:
🙏 thanks @AndreaGiardini

PR #319, opened 2021-05-17
update TiTiler dependencies

vincentsarago:
this pr does:
- update rio-tiler version
- update `rescale` query-parameter
```
# before 
# previously, rio-tiler was splitting a list of input range in tuple of 2
rescale=0,1000,0,1000,0,1000

# now
# rio-tiler 2.1 now expect sequence of tuple in form of Sequence[Tuple[Num, Num]]
rescale=0,1000&rescale=0,1000&rescale=0,1000
```

FYI, I'm following best practices for `list` as shown in https://fastapi.tiangolo.com/tutorial/query-params-str-validations/?h=list#query-parameter-list-multiple-values

- update cogeo-mosaic dep to 3.0
- update fastapi to 0.65


CHANGES.md
```diff
@@ -2,6 +2,10 @@
 
 ## 0.3.2 (TBD)
 
+### titiler.core
+
+* update rio-tiler dependency to `2.1` version and update `rescale` query-parameter (https://github.com/developmentseed/titiler/issues/319)
+
```
vincentsarago, line 8: TODO

titiler/application/setup.py
```diff
@@ -6,6 +6,7 @@
     long_description = f.read()
 
 inst_reqs = [
+    "rio-cogeo>=2.2,<2.3",
```
vincentsarago, line 9: I don't think we need `,<2.3` here, we use few rio-cogeo function and I don't see how it will be breaking in 2.3

titiler/application/titiler/application/templates/cog_index.html
```diff
@@ -432,16 +432,17 @@
...
-                params.rescale = `${min1},${max1},${min2},${max2},${min3},${max3}`
+                const r_rescale = `${scope.config[r].min},${scope.config[r].max}`
+                const g_rescale = `${scope.config[g].min},${scope.config[g].max}`
+                const b_rescale = `${scope.config[b].min},${scope.config[b].max}`
+                const rescale_params = [r_rescale, g_rescale, b_rescale].map(i => `rescale=${i}`).join('&')
+                console.log(rescale_params)
+                url_params += `&${rescale_params}`
             }
-            const url_params = Object.keys(params).map(i => `${i}=${params[i]}`).join('&')
+            console.log(url_params)
```
vincentsarago, line 445: TODO remove

titiler/core/setup.py
```diff
@@ -6,15 +6,14 @@
...
+    "fastapi==0.65.1",
     "geojson-pydantic",
     "jinja2>=2.11.2,<3.0.0",
     "morecantile>=2.1,<2.2",
     "numpy",
     "pydantic",
     "rasterio",
-    "rio-cogeo>=2.1,<2.2",
-    "rio-tiler>=2.0.5,<2.1",
+    "rio-tiler>=2.1,<2.2",
```
vincentsarago, line 16: should we pin rio-tiler ?

titiler/core/setup.py
```diff
@@ -6,15 +6,14 @@
     long_description = f.read()
 
 inst_reqs = [
-    "fastapi==0.63.0",
+    "fastapi==0.65.1",
```
vincentsarago, line 9: do we need to pin fastapi ?

titiler/core/titiler/core/factory.py
```diff
@@ -410,14 +410,20 @@ def tilejson(
...
+            qs_key_to_remove = [
+                "TileMatrixSetIt",
+                "tile_format",
+                "tile_scale",
+                "minzoom",
+                "maxzoom",
+            ]
+            qs = [
+                (key, value)
+                for (key, value) in request.query_params._list
```
vincentsarago, line 422: `dict(request.query_params)` will not keep the multiple inputs for `rescale` so we need to use `._list`.

titiler/core/titiler/core/resources/enums.py
```diff
@@ -22,6 +22,8 @@ class MediaType(str, Enum):
     geojson = "application/geo+json"
     html = "text/html"
     text = "text/plain"
+    pbf = "application/x-protobuf"
+    mvt = "application/x-protobuf"
```
vincentsarago, line 26: not related to this PR but might be useful in near future!

titiler/core/setup.py
kylebarron: IMO `titiler.core` shouldn't strictly pin fastapi. It's still worthwhile to provide a range if you know you use specific features, but I think the _application_ should be the one strictly pinning dependencies.

titiler/core/setup.py
kylebarron: Likewise, IMO a range is ideal for `titiler.core`.

kylebarron (approved):

titiler/core/setup.py
```diff
@@ -6,15 +6,14 @@
     long_description = f.read()
 
 inst_reqs = [
-    "fastapi==0.63.0",
+    "fastapi>=0.65,<0.66",
     "geojson-pydantic",
     "jinja2>=2.11.2,<3.0.0",
-    "morecantile>=2.1,<2.2",
+    "morecantile",
```
vincentsarago, line 12: morecantile version is already well defined in rio-tiler.

titiler/core/setup.py
```diff
@@ -6,15 +6,14 @@
     long_description = f.read()
 
 inst_reqs = [
-    "fastapi==0.63.0",
+    "fastapi>=0.65,<0.66",
```
vincentsarago, line 9: https://fastapi.tiangolo.com/deployment/versions/#about-fastapi-versions

PR #320, opened 2021-05-28
fix: update instructions for local install to make all packages mutable

philvarner:
update instructions for local install to make all packages mutable (-e), general install documentation improvements

The existing directions for running locally with uvicorn only applied mutability `-e` on titiler/core and not application or mosaic, so only changes to core were picked up by reload, even through changes in application or mosaic still triggered a reload -- which was very confusing!


vincentsarago:
🙏 thanks @philvarner

PR #321, opened 2021-06-03
Case insensitive query parameters

lorenzori:
Hi, thank you for this awesome tool! 

Some client applications require URL query parameters to be case-insensitive, so I have added a middleware to make all query parameters in the request lowercase following [an example from an issue in FastAPI](https://github.com/tiangolo/fastapi/issues/826). 

With this PR for example both `https://host/cog/WMTSCapabilities.xml?url=path/to/cog/...tif` and `https://host/cog/WMTSCapabilities.xml?URL=path/to/cog/...tif`
 will work.

I am not 100% the WMTS standard says anything about this so happy with any feedback.

### DIFF
--- a/titiler/application/titiler/application/main.py
+++ b/titiler/application/titiler/application/main.py
@@ -9,6 +9,7 @@ from titiler.application.middleware import (
     CacheControlMiddleware,
     LoggerMiddleware,
     TotalTimeMiddleware,
+    CaseInsensitiveMiddleware
 )
 from titiler.application.routers import cog, mosaic, stac, tms
 from titiler.application.settings import ApiSettings
@@ -66,6 +67,8 @@ app.add_middleware(
     cachecontrol=api_settings.cachecontrol,
     exclude_path={r"/healthz"},
 )
+
+app.add_middleware(CaseInsensitiveMiddleware)
 
 if api_settings.debug:
     app.add_middleware(LoggerMiddleware, headers=True, querystrings=True)
--- a/titiler/application/titiler/application/middleware.py
+++ b/titiler/application/titiler/application/middleware.py
@@ -88,3 +88,21 @@ class LoggerMiddleware(BaseHTTPMiddleware):
 
         response = await call_next(request)
         return response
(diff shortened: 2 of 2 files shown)


titiler/application/titiler/application/middleware.py
```diff
@@ -88,3 +88,18 @@ async def dispatch(self, request: Request, call_next):
 
         response = await call_next(request)
         return response
+
+
+class CaseInsensitiveMiddleware(BaseHTTPMiddleware):
```
vincentsarago, line 93: what do you think of `LowerCaseQueryStringMiddleware(BaseHTTPMiddleware)`?

titiler/application/titiler/application/main.py
```diff
@@ -67,6 +68,8 @@
     exclude_path={r"/healthz"},
 )
 
+app.add_middleware(CaseInsensitiveMiddleware)
```
vincentsarago, line 71: I'm kinda hesitant to add this as default.

vincentsarago:
🙏 thanks for the PR @lorenzori 

As I said in the comment, I'm not 💯 sure we want to add this middleware by default but I think it's fine to add it in the code so people can add it to their own application. 

The `titiler.application` module aims to be a demo and we highly suggest that titiler users create their own application.  

I opened another issue https://github.com/developmentseed/titiler/issues/322 because mixing lowercase/uppercase will break the tilejson and wmts endpoints.

lorenzori:
hey @vincentsarago thanks for the feedback! I have made the `LowerCaseQueryStringMiddleware` middleware optional as suggested using flag `TITILER_API_LOWER_CASE_QUERY_PARAMETERS`.

We have looked into making only the query keys lowercase instead of the entire query to avoid breaking the other endpoints and tests pass now locally, however is a bit of a hack as I didn't find an elegant solution with `Starlette`'s `Request`.

vincentsarago:
🙏 thanks @lorenzori, I added tests and fixed some format errors

PR #330, opened 2021-06-11
limit AWS S3 access

### LINKED ISSUE #329: Lambda S3 bucket permissions are too permissive
#### Problem description

This describes a code issue and an example issue related to granting s3 bucket permissions to the lambda.

The examples for giving bucket permissions look like:

```
TITILER_STACK_BUCKETS='["my-bucket*", "*"]'
```

The code that creates the permissions looks like:

```
if settings.buckets:
            perms.append(
                iam.PolicyStatement(
                    actions=["s3:GetObject", "s3:HeadObject"],
                    resources=[f"arn:aws:s3:::{bucket}*" for bucket in settings.buckets],
                )
            )
```

One issue is that the example should be `my-bucket` rather than `my-bucket*`, since that ends up with an s3 permission on `arn:aws:s3:::my-bucket**` and ``arn:aws:s3:::**` (double wildcard), which is no different than a single *. Most users likely want to only grant access to the exactly-named bucket, not all buckets prefixed with that value, so I think it is a misleading example.

However, even with a TITILER_STACK_BUCKETS that only has `my-bucket`, I think the resource URN generated for the permissions should have a `/` between the bucket and *, e.g., `arn:aws:s3:::{bucket}/*`, because `arn:aws:s3:::{bucket}*` gives permissions to *any* bucket that is prefixed with the TITILER_STACK_BUCKETS value, not just all the objects in the one named buckets. 

I could imagine a scenario where a user had buckets `foo` and `foo-private`, and putting just `foo` in TITILER_STACK_BUCKETS is inadvertently granting access to `foo-private`. 

#### Expected Output

A CFN template that only has exactly the access to the buckets specified.

#### Environment Information

Os, Python version, ...

vincentsarago:
closes #329 

cc @philvarner

### DIFF
--- a/deployment/aws/cdk/app.py
+++ b/deployment/aws/cdk/app.py
@@ -169,7 +169,7 @@ perms = []
 if settings.buckets:
     perms.append(
         iam.PolicyStatement(
-            actions=["s3:GetObject", "s3:HeadObject"],
+            actions=["s3:GetObject"],
             resources=[
                 f"arn:aws:s3:::{bucket}/{settings.key}" for bucket in settings.buckets
             ],
--- a/deployment/aws/cdk/app.py
+++ b/deployment/aws/cdk/app.py
@@ -170,7 +170,9 @@ if settings.buckets:
     perms.append(
         iam.PolicyStatement(
             actions=["s3:GetObject", "s3:HeadObject"],
-            resources=[f"arn:aws:s3:::{bucket}*" for bucket in settings.buckets],
+            resources=[
+                f"arn:aws:s3:::{bucket}/{settings.key}" for bucket in settings.buckets
+            ],
         )
     )
 
--- a/deployment/aws/cdk/config.py
+++ b/deployment/aws/cdk/config.py
@@ -32,6 +32,7 @@ class StackSettings(pydantic.BaseSettings):
     # S3 bucket names where TiTiler could do HEAD and GET Requests
     # specific private and public buckets MUST be added if you want to use s3:// urls
     # You can whitelist all bucket by setting `*`.
+    # ref: https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html
     buckets: List = []
 
     # S3 key pattern to limit the access to specific items (e.g: "my_data/*.tif")
--- a/deployment/aws/cdk/config.py
+++ b/deployment/aws/cdk/config.py
@@ -29,10 +29,13 @@ class StackSettings(pydantic.BaseSettings):
         "VSI_CACHE_SIZE": "5000000",  # 5 MB (per file-handle)
     }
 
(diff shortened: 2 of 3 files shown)


philvarner:
👍🏻

philvarner:
The examples how how to use `TITILER_STACK_BUCKETS` also need to be updated to remove the `*` at the end of them and add an example of using `key`.

deployment/aws/cdk/app.py
```diff
@@ -170,7 +170,9 @@ def __init__(
     perms.append(
         iam.PolicyStatement(
             actions=["s3:GetObject", "s3:HeadObject"],
-            resources=[f"arn:aws:s3:::{bucket}*" for bucket in settings.buckets],
+            resources=[
```
philvarner, line 173: this should have a check for if `bucket == "*"` and not append the key in that case.

philvarner:
Ah, here's the documentation https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html under Wildcards

you can use multiple `?` and `*` for the key name.

deployment/aws/cdk/app.py
vincentsarago: > otherwise you end up with */*

I'm not sure to understand, reading the documentation is seems to be that `*/*` is valid

philvarner: It's valid, but it's just redundant to `*`, but maybe it's fine if it makes the code simpler.

deployment/aws/cdk/config.py
```diff
@@ -29,11 +29,15 @@ class StackSettings(pydantic.BaseSettings):
...
 
-    # add S3 bucket where TiTiler could do HEAD and GET Requests
+    # S3 bucket names where TiTiler could do HEAD and GET Requests
     # specific private and public buckets MUST be added if you want to use s3:// urls
     # You can whitelist all bucket by setting `*`.
+    # ref: https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html
     buckets: List = []
 
+    # S3 key pattern to limit the access to specific items (e.g: "my_data/*.tif")
+    key: str = "*"
```
vincentsarago, line 39: status: I tried and it worked!

deployment/aws/cdk/app.py
```diff
@@ -169,8 +169,10 @@ def __init__(
 if settings.buckets:
     perms.append(
         iam.PolicyStatement(
-            actions=["s3:GetObject", "s3:HeadObject"],
-            resources=[f"arn:aws:s3:::{bucket}*" for bucket in settings.buckets],
+            actions=["s3:GetObject"],
```
vincentsarago, line 172: `HeadObject` doesn't exist .... (maybe it never did 🤷‍♂️)

PR #332, opened 2021-06-22
Signed url example

vincentsarago:
ref: https://github.com/developmentseed/titiler/discussions/331

### DIFF
--- a/docs/examples/code/working_with_signed_urls.md
+++ b/docs/examples/code/working_with_signed_urls.md
@@ -44,6 +44,17 @@ def DatasetPathParams(
         url = base64.b64decode(url).decode()
     return url
 
+# Another solution is to pass only the signed url path parameters encoded in base64
+# def DatasetPathParams(
+#     url: str = Query(..., description="Dataset URL"),
+#     url_params: str = Query(
+#         None, description="Base64 encoded Query parameters to add to the dataset URL."
+#     ),
+# ) -> str:
+#     """DatasetPath Params."""
+#     if url_params:
+#         url += f"{b64decode(url_params).decode()}"
+#     return url
 
 app = FastAPI(title="My simple app")
 
--- a/docs/examples/code/working_with_signed_urls.md
+++ b/docs/examples/code/working_with_signed_urls.md
@@ -1,0 +1,124 @@
+
+**ref**: https://github.com/developmentseed/titiler/discussions/331
+
+**Goal**: allow users to pass signed url
+
+**requirements**: titiler.core
+
+## What / Why / How
+
+Passing a signed URL for a dataset is not supported by default in TiTiler because the parameters from the signed url are conflicting with the query parameters from the application itself. In order to allow signed url in the application there are two solutions:
+
+
+### 1. URL Encoding
+
+The easiest way (from the application POV) to allow complex URL, is to allow encoded url as input parameter.
+
+```python
+import base64
+url = "http://my.dataset.com/cog.tif?p=aaa&c&1234&t=4321"
+
+# base64.b64encode(url.encode())
+>>> "aHR0cDovL215LmRhdGFzZXQuY29tL2NvZy50aWY/cD1hYWEmYyYxMjM0JnQ9NDMyMQ=="
+```
+
+When base64 encoded, url is just a regular string and thus will be valid in the application.
+
+```python
+"""Minimal COG tiler with Signed URL support."""
+
+import base64
+from titiler.core.factory import TilerFactory
+from titiler.core.errors import DEFAULT_STATUS_CODES, add_exception_handlers
+
+from fastapi import FastAPI
+
+
+# Custom Path dependency which can `decode` a base64 url
+def DatasetPathParams(
+    url: str = Query(..., description="Dataset URL"),
+    base64_encoded: bool = Query(None)
+) -> str:
+    """Create dataset path from args"""
+    if base64_encoded:
+        url = base64.b64decode(url).decode()
+    return url
+
+
+app = FastAPI(title="My simple app")
+
+cog = TilerFactory(path_dependency=DatasetPathParams)
+app.include_router(cog.router, tags=["Cloud Optimized GeoTIFF"])
+
+add_exception_handlers(app, DEFAULT_STATUS_CODES)
+
+
+@app.get("/healthz", description="Health Check", tags=["Health Check"])
+def ping():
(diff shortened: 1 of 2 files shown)


vincentsarago:
FYI: CI is failing because of https://github.com/developmentseed/cogeo-mosaic/issues/177

docs/examples/code/working_with_signed_urls.md
```diff
@@ -0,0 +1,135 @@
+
```
kylebarron, line 1: Could probably use a top-level header?

```
# Example with Signed URLs
```

docs/examples/code/working_with_signed_urls.md
```diff
@@ -0,0 +1,135 @@
...
+**requirements**: titiler.core
+
+## What / Why / How
+
+Passing a signed URL for a dataset is not supported by default in TiTiler because the parameters from the signed url are conflicting with the query parameters from the application itself. In order to allow signed url in the application there are two solutions:
+
+
+### 1. URL Encoding
+
+The easiest way (from the application POV) to allow complex URL, is to allow encoded url as input parameter.
```
kylebarron, line 15:
```suggestion
The easiest way (from the application's point of view) to allow complex URLs is to allow an encoded url as an input parameter.
```

docs/examples/code/working_with_signed_urls.md
```diff
@@ -0,0 +1,135 @@
+
+**ref**: https://github.com/developmentseed/titiler/discussions/331
+
+**Goal**: allow users to pass signed url
+
+**requirements**: titiler.core
+
+## What / Why / How
+
+Passing a signed URL for a dataset is not supported by default in TiTiler because the parameters from the signed url are conflicting with the query parameters from the application itself. In order to allow signed url in the application there are two solutions:
```
kylebarron, line 10:
```suggestion
Passing a signed URL for a dataset is not supported by default in TiTiler because the parameters from the signed url conflict with the query parameters from the application itself. In order to allow signed url in the application there are two solutions:
```

docs/examples/code/working_with_signed_urls.md
```diff
@@ -0,0 +1,135 @@
...
+
+```python
+import base64
+url = "http://my.dataset.com/cog.tif?p=aaa&c&1234&t=4321"
+
+# base64.b64encode(url.encode())
+>>> "aHR0cDovL215LmRhdGFzZXQuY29tL2NvZy50aWY/cD1hYWEmYyYxMjM0JnQ9NDMyMQ=="
+```
+
+When base64 encoded, url is just a regular string and thus will be valid in the application.
```
kylebarron, line 25:
```suggestion
When base64 encoded, the url is just a regular string and thus will be valid in the application.
```

kylebarron: Maybe mention it's the provided `&` characters that mess things up?

docs/examples/code/working_with_signed_urls.md
```diff
@@ -0,0 +1,135 @@
...
+def DatasetPathParams(
+    url: str = Query(..., description="Dataset URL"),
+    base64_encoded: bool = Query(None)
+) -> str:
+    """Create dataset path from args"""
+    if base64_encoded:
+        url = base64.b64decode(url).decode()
+    return url
+
+# Another solution is to pass only the signed url path parameters encoded in base64
```
kylebarron, line 47: Nit: Instead of keeping as a comment, I'd suggest having two separate code blocks with prose around them, and then a third block that uses `DatasetPathParams`

`````md

Create a PathParams dependency to pass to the tiler, two approaches

```py
# first approach
def DatasetPathParams
```

```py
# second approach
def DatasetPathParams
```

`````

File: docs/examples/code/working_with_signed_urls.md [hunk 6]
```diff
@@ -0,0 +1,135 @@
...
+
+
+### 2. Signing URL in the application
+
+Another solution is to sign the URL directly in TiTiler.
+
+```python
+"""Minimal COG tiler with Signed URL support."""
+
+import my_provider  # e.g AWS, Google, ...
```
kylebarron, line 96: Since the example is abstract for AWS and Google, but concrete for Microsoft's Planetary computer, I think it would make sense to make this example usable for PC and then mention it could be adapted for AWS/Google as well

docs/examples/code/working_with_signed_urls.md
```diff
@@ -0,0 +1,135 @@
...
+
+from titiler.core.factory import TilerFactory
+from titiler.core.errors import DEFAULT_STATUS_CODES, add_exception_handlers
+
+from fastapi import FastAPI
+
+
+# Custom Path dependency which will sign url
+# !!! You may want to add caching here to avoid to many call to the signing provider !!!
```
kylebarron, line 105: Maybe add a note that caching can be helpful to reduce load but could also present some hard-to-debug errors if, for example, each signed link has an expiration time and that isn't communicated to the application, so the application uses cached, expired signing tokens.

mkdocs.yml
```diff
@@ -45,6 +45,7 @@ nav:
       - Tiler with Cache: "examples/code/tiler_with_cache.md"
       - Mosaic from COG urls: "examples/code/mosaic_from_urls.md"
       - Tiler with custom Colormap dependency: "examples/code/tiler_with_custom_colormap.md"
+      - Working with signed URL: "examples/code/working_with_signed_urls.md"
```
kylebarron, line 48:
```suggestion
      - Loading data with signed URLs: "examples/code/working_with_signed_urls.md"
```

kylebarron (approved):

docs/examples/code/working_with_signed_urls.md
vincentsarago: I'm not sure to understand @kylebarron

kylebarron: This document (in isolation) doesn't have a title. Maybe mkdocs adds a title if you don't provide an initial `h1` header in the markdown?

vincentsarago: yeah, mkdocs adds a title. I think it's fine as it is

docs/examples/code/working_with_signed_urls.md
vincentsarago: I won't add more things here. I assume people working with signed URL know this.

PR #343, opened 2021-07-22
Add ROOT_PATH env variable to simplify running Titiler behind a proxy

tayden:
I'd really like to get Titiler running on an existing server, and have requests to the root path `/titiler` passed to a running docker container running on, e.g. port 9000. However, there's currently no way to specify the root_path parameter in the FastAPI instantiation, so the app has been required to run at the root location `/`. (see FastAPI docs on [running behind a proxy](https://fastapi.tiangolo.com/advanced/behind-a-proxy/#setting-the-root_path-in-the-fastapi-app))

This PR allows users to set the root_path parameter of the FastAPI instance via an environment variable so that it's dead simple to run the service behind a reverse proxy. It defaults that root_path location to "/", so nothing anybody currently has running breaks.

I suspect that for this to be accepted, some documentation has to be written up for this behaviour. I'm just not sure where that should go. If someone wants to tell me where to put that, I can fill in those details.

### DIFF
--- a/src/titiler/application/titiler/application/settings.py
+++ b/src/titiler/application/titiler/application/settings.py
@@ -1,4 +1,5 @@
 """Titiler API settings."""
+import os
 
 import pydantic
 
@@ -9,6 +10,7 @@ class ApiSettings(pydantic.BaseSettings):
     name: str = "titiler"
     cors_origins: str = "*"
     cachecontrol: str = "public, max-age=3600"
+    root_path: str = os.getenv("ROOT_PATH", "")
     debug: bool = False
 
     disable_cog: bool = False
--- a/src/titiler/application/titiler/application/settings.py
+++ b/src/titiler/application/titiler/application/settings.py
@@ -1,5 +1,4 @@
 """Titiler API settings."""
-import os
 
 import pydantic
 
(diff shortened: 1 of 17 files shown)


vincentsarago:
@tayden thanks for the PR

Would it make more sense to add this to https://github.com/developmentseed/titiler/blob/master/src/titiler/application/titiler/application/settings.py#L6 ?

tayden:
@vincentsarago That makes sense to me. I've updated the code for both of those suggestions

src/titiler/application/titiler/application/settings.py
```diff
@@ -9,6 +10,7 @@ class ApiSettings(pydantic.BaseSettings):
     name: str = "titiler"
     cors_origins: str = "*"
     cachecontrol: str = "public, max-age=3600"
+    root_path: str = os.getenv("ROOT_PATH", "")
```
vincentsarago, line 13: you don't need to use os.getenv here (see https://pydantic-docs.helpmanual.io/usage/settings/#environment-variable-names) 

this means that to set `root_path` you'll need to have `TITILER_API_ROOT_PATH` either in your environment or in a `.env` file

vincentsarago (changes requested):

src/titiler/application/titiler/application/settings.py
tayden: Oh! Cool, ok, will change. Thanks for the help!

vincentsarago:
excellent, thanks @tayden 

Tests are failing but it's unrelated to the changes (and have been fixed on the master branch just now). Could you merge from master and add a note in the changelog 🙏

vincentsarago (approved):

vincentsarago:
🙏 thanks @tayden

PR #346, opened 2021-07-27
Deploy docker image directly to lambda

aripollak:
I noticed that the existing code was causing CDK to build the docker image every time you invoke `cdk`, even for something like `cdk ls`. This PR changes the lambda to use the built docker image directly instead of the Python runtime, and that seems to be intelligent enough to only rebuild the Docker image when Dockerfile or handler.py changes, and only on `deploy`.

### DIFF
--- a/deployment/aws/lambda/Dockerfile
+++ b/deployment/aws/lambda/Dockerfile
@@ -1,15 +1,7 @@
-FROM lambci/lambda:build-python3.8
+FROM public.ecr.aws/lambda/python:3.8
 
-WORKDIR /tmp
+RUN pip install titiler.application mangum>=0.10.0
 
-RUN pip install titiler.application mangum>=0.10.0 -t /var/task --no-binary numpy,pydantic
+COPY lambda/handler.py ${LAMBDA_TASK_ROOT}/
 
-# Reduce package size and remove useless files
-RUN cd /var/task && find . -type f -name '*.pyc' | while read f; do n=$(echo $f | sed 's/__pycache__\///' | sed 's/.cpython-[2-3][0-9]//'); cp $f $n; done;
-RUN cd /var/task && find . -type d -a -name '__pycache__' -print0 | xargs -0 rm -rf
-RUN cd /var/task && find . -type f -a -name '*.py' -print0 | xargs -0 rm -f
-RUN find /var/task -type d -a -name 'tests' -print0 | xargs -0 rm -rf
-RUN rm -rdf /var/task/numpy/doc/
-RUN rm -rdf /var/task/uvicorn
-
-COPY lambda/handler.py /var/task/handler.py
+CMD ["handler.handler"]
--- a/deployment/aws/lambda/Dockerfile
+++ b/deployment/aws/lambda/Dockerfile
@@ -2,6 +2,6 @@ FROM public.ecr.aws/lambda/python:3.8
 
 RUN pip install titiler.application mangum>=0.10.0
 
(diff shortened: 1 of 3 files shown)


vincentsarago:
@aripollak Thanks for the PR 🙏 
I'm not 100% to understand what this does though. 

>  This PR changes the lambda to use the built docker image directly

Does this means we are not creating a lambda package anymore but using docker in aws lambda? 

If that the case, I'm not sure I want to merge this. Last time I tried using docker image for lambda I experienced long  startup which meant the lambda wasn't really scaling fast enough for my application. Maybe it changed 🤷.

But I feel you buiding the docker image everytime is not great at all but I blame cdk for this ;-)

deployment/aws/lambda/Dockerfile
```diff
@@ -1,15 +1,7 @@
...
+RUN pip install titiler.application mangum>=0.10.0
 
-RUN pip install titiler.application mangum>=0.10.0 -t /var/task --no-binary numpy,pydantic
+COPY lambda/handler.py ${LAMBDA_TASK_ROOT}/
 
-# Reduce package size and remove useless files
-RUN cd /var/task && find . -type f -name '*.pyc' | while read f; do n=$(echo $f | sed 's/__pycache__\///' | sed 's/.cpython-[2-3][0-9]//'); cp $f $n; done;
-RUN cd /var/task && find . -type d -a -name '__pycache__' -print0 | xargs -0 rm -rf
-RUN cd /var/task && find . -type f -a -name '*.py' -print0 | xargs -0 rm -f
-RUN find /var/task -type d -a -name 'tests' -print0 | xargs -0 rm -rf
```
vincentsarago, line 11: This could still be kept, In addition of removing useless files it make startup faster!

geospatial-jeff:
> If that the case, I'm not sure I want to merge this. Last time I tried using docker image for lambda I experienced long startup which meant the lambda wasn't really scaling fast enough for my application. Maybe it changed 🤷.

It's true that using a container runtime for lambda causes longer cold starts because its more expensive to pull from ECR, especially when the container contains big dependencies like GDAL.

deployment/aws/lambda/Dockerfile
aripollak: To reduce the docker image size, we'd have to do it in the same RUN as the pip install, but I'd want to see how much space/time it saves for this.

vincentsarago:
FYI: if you are looking for a really fast way to deploy TiTiler lambda, you could check https://github.com/lambgeo/titiler-layer

aripollak:
Confirmed that a request after a cold start using this mechanism takes about 7-8 seconds (which is far higher than some of the benchmarks I've seen elsewhere). For my use case I think that's a tradeoff I'm willing to make for now since Titiler is being deployed as part of a larger stack, but if it becomes a problem then adding Cloudfront or switching to ECS could be potential options.
I haven't tried other bundling options from https://docs.aws.amazon.com/cdk/api/latest/python/aws_cdk.aws_lambda/README.html#bundling-asset-code, but it's possible they would also avoid the delay. Note that the existing code package is already getting pretty close to the Lambda limit - 219 MB out of 250 MB - whereas the Docker image size is 973 MB out of a 10 GB limit.

vincentsarago:
I'm sorry @aripollak but I cannot accept this PR as it is because the lambda would become much slower. Worth noting that the CDK code is quite old and deserve an update (e.g we could use `aws_lambda.Code.from_docker_build`, which seems better than what we have right now).

PR #347, opened 2021-07-27
add statistics endpoints

### LINKED ISSUE #340: add `/statistics` endpoints
We already have `metadata` endpoints which return basic statistics (percentile/min/max/stdev and histogram but I think we would be nice to have more advances statistics a bit like https://github.com/perrygeo/python-rasterstats/blob/master/src/rasterstats/main.py#L223-L258

idea is to add 2 endpoints: 
- GET /statistics: return stats for the whole dataset
- POST /statistics: return stats for a geojson feature covering the dataset

vincentsarago:
closes #340 

This PR adds 2 new endpoints per TilerFactories
`GET - /statistics`: get stats from a preview of a dataset
`POST - /statistics`: get stats from a feature covering a dataset

### To Be Discussed

Per the HTML specification, `list` must be passed using multiple parameter (e.g `p=1&p=2`) and is also supported in openapi. When we started titiler we (I) didn't knew about this and used `comma delimited string` (e.g `bidx=1,2,3`) to allow list input. In the PR the statistics endpoints have 2 `list` inputs: **c** (categories) and **p** (percentiles). I've choose to keep the name small because those will be repeated in the query_param (e.g `p=1&p=2` to define `percentiles=[1, 2]`).

I'm open to discuss either we should follow the spec or use `comma delimited values`. I've already raised this over https://github.com/developmentseed/titiler/pull/262#issuecomment-792863831

IMO we should switch every `list` input to this, but as it's a breaking change I'll think starting to implement this for new QueryParameters is the safest way 🤷 

Here is example of `statistics` result
```
$ curl http://127.0.0.1:8000/cog/statistics\?url\=https://rio-tiler-dev.s3.amazonaws.com/data/eu_webAligned_256px.tif&bidx=1
[
  {
    "min": 38,
    "max": 255,
    "mean": 125.75282900811122,
    "count": 262106,
    "sum": 32960571,
    "std": 36.320596765212024,
    "median": 102,
    "percentile_2": 0,
    "percentile_98": 178,
    "valid_percent": 25
  }
]

$ curl http://127.0.0.1:8000/cog/statistics\?url\=https://rio-tiler-dev.s3.amazonaws.com/data/eu_webAligned_256px.tif&bidx=1&categorical=true&c=1&c=2&c=3
[
  {
    "categories": {
      "1": 0,
      "2": 0,
      "3": 0
    },
    "valid_percent": 25
  }
]
```

### DIFF
--- a/src/titiler/core/titiler/core/factory.py
+++ b/src/titiler/core/titiler/core/factory.py
@@ -6,7 +6,7 @@ from typing import Any, Callable, Dict, List, Optional, Type, Union
 from urllib.parse import urlencode, urlparse
 
 import rasterio
-from geojson_pydantic.features import Feature
+from geojson_pydantic.features import Feature, FeatureCollection
 from morecantile import TileMatrixSet
 from rio_tiler.io import BaseReader, COGReader, MultiBandReader, MultiBaseReader
 from rio_tiler.models import Bounds, Info, Metadata
@@ -821,6 +821,9 @@ class TilerFactory(BaseTilerFactory):
 
         @self.router.post(
             "/statistics",
+            response_model=Union[Feature, FeatureCollection],
+            response_model_exclude_none=True,
+            response_class=GeoJSONResponse,
             responses={
                 200: {
                     "content": {"application/json": {}},
@@ -829,7 +832,9 @@ class TilerFactory(BaseTilerFactory):
             },
         )
         def geojson_statistics(
-            feature: Feature = Body(..., descriptiom="GeoJSON Feature."),
+            features: Union[FeatureCollection, Feature] = Body(
+                ..., descriptiom="GeoJSON Feature or FeatureCollection."
+            ),
             src_path=Depends(self.path_dependency),
             layer_params=Depends(self.layer_dependency),
             image_params=Depends(self.img_dependency),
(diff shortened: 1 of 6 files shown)


src/titiler/core/titiler/core/factory.py
```diff
@@ -772,6 +776,88 @@ def geojson_crop(
...
+                        feature.dict(exclude_none=True),
+                        **layer_params.kwargs,
+                        **image_params.kwargs,
+                        **dataset_params.kwargs,
+                        **kwargs,
+                    ).as_masked()
+
+            return data_stats(
+                data, categorical=categorical, categories=c, percentiles=p,
+            )
```
vincentsarago, line 859: We should return a geojson with statistics in the feature properties

src/titiler/core/titiler/core/factory.py
```diff
@@ -772,6 +776,88 @@ def geojson_crop(
...
+            "/statistics",
+            responses={
+                200: {
+                    "content": {"application/json": {}},
+                    "description": "Return dataset's statistics.",
+                }
+            },
+        )
+        def geojson_statistics(
+            feature: Feature = Body(..., descriptiom="GeoJSON Feature."),
```
vincentsarago, line 832: We could allow Feature and Feature collection (and loop over the features)🤷‍♂️

PR #352, opened 2021-08-06
Assets for tile

vincentsarago:
ref https://github.com/developmentseed/titiler/discussions/350#discussioncomment-1138576

IMO it's better to use `tile` instead of quadkey, then we could in theory support tile in different TMS (which might not support quadkeys)

cc @mackdelany


mackdelany:
Good call @vincentsarago, I agree!

PR #374, opened 2021-09-22
use simplejson to allow NaN/inf/-inf in JSON response

vincentsarago:
ref https://github.com/developmentseed/titiler/discussions/373 and https://github.com/tiangolo/fastapi/issues/3417

By default FastAPI/starlette will raise an error when we try to JSON encode `float('nan')` or `+/-float('inf')`.

Recently in rio-tiler we added `nodata_value` in `rio_tiler.models.Info` which can sometimes be set to Nan/inf/-inf. But because of FastAPI/starlette defaults, titiler was failing to return response for /info and /metadata. 

This PR adds a custom `JSONResponse` using the great `simplejson` (which will translate NaN/inf/-inf values to null).

Another solution would have been to change rio-tiler behaviour to stringify nodata value but in my opinion this is an application issues not rio-tiler's


src/titiler/core/titiler/core/resources/responses.py
```diff
@@ -1,14 +1,36 @@
...
+        """Render JSON.
+
+        Same defaults as starlette.responses.JSONResponse.render but allow NaN to be replaced by null using simplejson
+        """
+        return json.dumps(
+            content,
+            ensure_ascii=False,
+            allow_nan=False,
+            indent=None,
+            ignore_nan=True,
```
vincentsarago, line 29: ✨

vincentsarago:
Note: in the last commit I choose to use the custom JSONResponse directly in the tiler factories directly to kinda `force` the response. I think this is safer choice.

PR #406, opened 2021-11-23
add method to add more dependencies to specific endpoints

vincentsarago:
based on https://github.com/stac-utils/stac-fastapi/pull/295

cc @alukach


src/titiler/core/titiler/core/factory.py
```diff
@@ -166,6 +177,32 @@ def url_for(self, request: Request, name: str, **path_params: Any) -> str:
...
+        for route in self.router.routes:
+            for scope in scopes:
+                match, _ = route.matches({"type": "http", **scope})
+                if match != Match.FULL:
+                    continue
+
+                # Mimicking how APIRoute handles dependencies:
+                # https://github.com/tiangolo/fastapi/blob/1760da0efa55585c19835d81afa8ca386036c325/fastapi/routing.py#L408-L412
+                for depends in dependencies[::-1]:
+                    route.dependant.dependencies.insert(  # type: ignore
```
vincentsarago, line 199: here we add the dependency

src/titiler/core/titiler/core/factory.py
```diff
@@ -166,6 +177,32 @@ def url_for(self, request: Request, name: str, **path_params: Any) -> str:
...
+                        0,
+                        get_parameterless_sub_dependant(
+                            depends=depends, path=route.path_format  # type: ignore
+                        ),
+                    )
+
```
vincentsarago, line 205: I confirmed here that the dependency was well inserted

vincentsarago:
This is not working as it is. While the dependencies are registered 

```python
from fastapi import FastAPI, security, Depends, HTTPException
from starlette import status

http_basic = security.HTTPBasic()


def must_be_bob(credentials: security.HTTPBasicCredentials = Depends(http_basic)):
    if credentials.username == 'bob':
        return True
    raise HTTPException(
        status_code=status.HTTP_401_UNAUTHORIZED,
        detail="You're not Bob",
        headers={"WWW-Authenticate": "Basic"},
    )

cog = TilerFactory(
    route_dependencies=[
        (
            [
                {"path": "/bounds", "method": "GET"},
                {"path": "/tiles/{z}/{x}/{y}", "method": "GET"},
            ],
            [Depends(must_be_bob)],
        ),
    ],
)
print(cog.router.routes[0].dependant.dependencies)
app = FastAPI()
app.include_router(cog.router, prefix="/something")
print(app.routes[4].dependant.dependencies)  # This is the `/bounds` endpoint

[<fastapi.dependencies.models.Dependant object at 0x137d06d30>, <fastapi.dependencies.models.Dependant object at 0x137d06d60>]
[<fastapi.dependencies.models.Dependant object at 0x137ee0940>]
```

src/titiler/core/titiler/core/factory.py
```diff
@@ -166,6 +177,32 @@ def url_for(self, request: Request, name: str, **path_params: Any) -> str:
...
+        self,
+        scopes: List[EndpointScope],
+        dependencies=List[params.Depends],
+    ):
+        """Add dependencies to routes.
+
+        Allows a developer to add dependencies to a route after the route has been defined.
+
+        """
+        for route in self.router.routes:
```
alukach, line 190:
```suggestion
        for route in self.app.router.routes:
```

This is the 🐛.  Once the api is established, I think we no longer care about `self.router`.

vincentsarago: there is no `app` in the Factory

src/titiler/core/titiler/core/factory.py
```diff
@@ -166,6 +177,32 @@ def url_for(self, request: Request, name: str, **path_params: Any) -> str:
...
+
+                # Mimicking how APIRoute handles dependencies:
+                # https://github.com/tiangolo/fastapi/blob/1760da0efa55585c19835d81afa8ca386036c325/fastapi/routing.py#L408-L412
+                for depends in dependencies[::-1]:
+                    route.dependant.dependencies.insert(  # type: ignore
+                        0,
+                        get_parameterless_sub_dependant(
+                            depends=depends, path=route.path_format  # type: ignore
+                        ),
+                    )
```
alukach, line 204: From https://github.com/stac-utils/stac-fastapi/pull/295/commits/99445db5a99db61273dd6993dffb500eee18d18d

```suggestion
                    )

            # Register dependency on route so that it isn't ignored if the routes
            # are added to an apt after this is route (e.g. app.include_router(router))
            # https://github.com/tiangolo/fastapi/blob/58ab733f19846b4875c5b79bfb1f4d1cb7f4823f/fastapi/routing.py#L677-L678
            route.dependencies.extend(dependencies)

```

vincentsarago:
😬 `TypedDict` is not available in python 3.7

PR #407, opened 2021-11-25
add simple uvicorn based docker image for K8s

vincentsarago:
As mentioned in https://fastapi.tiangolo.com/deployment/docker/#replication-number-of-processes, you'd better be served by *running a single Uvicorn process instead of running something like Gunicorn with Uvicorn workers* when using k8s. 

## ToDo

- do we need more configuration (e.g host, port, https://github.com/tiangolo/uvicorn-gunicorn-docker#advanced-usage)
- do we want to publish this image `ghcr.io/developmentseed/titiler-uvicorn:latest` 🤷‍♂️  

## Image sizes

- titiler: 1.39GB # official TiTiler image
- titiler-uvicorn: 621MB. # The one in the PR 😱 

@geospatial-jeff @emmanuelmathot any comments?

cc @geohacker @lossyrob


vincentsarago:
@emmanuelmathot @geospatial-jeff any comment?

PR #421, opened 2022-01-10
add Microsoft Azure function deployment

vincentsarago:
@robintw would you be able to review this? 🙏

### DIFF
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -2,4 +2,6 @@
 
 **Amazon Web Services** - [/aws](/deployment/aws)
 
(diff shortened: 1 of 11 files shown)


robintw (approved):
This looks great - thanks for adding it. It is almost exactly the same as the function definition I put together myself.

I haven't actually tried running this version, but I assume you've deployed it and it works.

vincentsarago:
@robintw I may have 👀  https://github.com/robintw/starlette-test-azure/blob/master/HttpTrigger1/function.json for inspiration 😉 

🙏

PR #438, opened 2022-02-24
Add missing type import

msfstef:
In one of the examples in the documentation a type import was missing


vincentsarago:
🙏 Thanks @msfstef

PR #444, opened 2022-03-03
Make http error cache control behavior configurable.

sharkinsspatial:
## What I am changing
While implementing edge caching for a titiler deployment we have a case where underlying source COGs are continually added in real time.  Currently a request that results in `404` `No assets found for tile` has the same `Cache-Control` header as a `200` response and is maintained by the edge cache for the same duration.  In our case we would prefer `404` responses to have a shorter duration in the edge cache so that requests are passed to the origin (where new assets may now be available).  This PR makes the upper range or http codes which include a `Cache-Control` header configurable so that edge caches can exert finer control on error response cache duration.

## How I did it
Extended `core/middleware` to support a configurable upper bound for response codes which include a `Cache-Control` header rather than the hardcoded `500` value.

## How you can test it
Via the project's `tox` testing entrypoint.

### DIFF
--- a/src/titiler/core/titiler/core/middleware.py
+++ b/src/titiler/core/titiler/core/middleware.py
@@ -19,6 +19,7 @@ class CacheControlMiddleware(BaseHTTPMiddleware):
         self,
         app: ASGIApp,
         cachecontrol: Optional[str] = None,
+        cachecontrol_http_code_range: Optional[int] = 500,
         exclude_path: Optional[Set[str]] = None,
     ) -> None:
         """Init Middleware.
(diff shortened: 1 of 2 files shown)


src/titiler/core/titiler/core/middleware.py
```diff
@@ -31,6 +32,7 @@ def __init__(
         """
         super().__init__(app)
         self.cachecontrol = cachecontrol
+        self.cachecontrol_http_code_range = cachecontrol_http_code_range
```
vincentsarago, line 35: I'm not sure about the name of the parameter. To me if we say `Range` I'm assuming the input to be a sequence (min, max), while here in reality we want the upper value of status_code.

what do you think about `cachecontrol_max_http_code` ?

sharkinsspatial: Strong agree 👍 .

vincentsarago (approved):
🙏 thanks @sharkinsspatial

sharkinsspatial:
I may revisit this a bit later @vincentsarago I'm still not fully sure of the Cloudfront behavior between `Cache-Control` and `Max TTL` for error responses but this should help diagnose. 🙇

PR #455, opened 2022-04-04
handle antimeridian and remove logger in exception handler

### LINKED ISSUE #453: remove `exc_info=True` in error logger
in https://github.com/developmentseed/titiler/blob/33d3492551aa6198446dc503842ee6c4e6a31332/src/titiler/core/titiler/core/errors.py#L60

we add `exc_info=True` which in theory should print the `the current exception information`, but in our case it's always returning `NoneType: None`.

My understanding is that we had exception handler for known exception which IMO do not require to log the exception information.

cc @geospatial-jeff

vincentsarago:
closes #453


src/titiler/core/titiler/core/errors.py
```diff
@@ -57,7 +54,6 @@ def exception_handler_factory(status_code: int) -> Callable:
     """
 
     def handler(request: Request, exc: Exception):
-        logger.error(exc, exc_info=True)
```
vincentsarago, line 60: IMO: error message will be forwarded to the client so we don't really need to `print` it

vincentsarago: @geospatial-jeff, I've decided to completely remove the logger because ☝️. If this is wrong I'll make the change you are proposing!

PR #464, opened 2022-05-06
Include all values for a query param in LowerCaseQueryStringMiddleware

samn:
## What I am changing

- Previously if multiple values for a query param were present on a
request (e.g. for a `List[T] = Query(...)` param) only one of the values
would be present after the `LowerCaseQueryStringMiddleware` ran.

## How I did it

- I noticed this issue in our production TiTiler deployment (running version 0.5.1). I copied the `LowerCaseQueryStringMiddleware` and made this change in our app to verify it worked as intended.
- Also added a unit test demonstrating the fix.

## How you can test it
- A unit test is included.


vincentsarago (commented):
🙏 thanks @samn

PR #478, opened 2022-05-17
Update __init__.py

Jnb2387:
version file was removed

## Available PR templates

- _Alternatively delete and start empty_


vincentsarago:
@Jnb2387 sorry about this 

we don't have to hardcode the version here but we can use 
```python
from titiler.core import __version__ as titiler_version
```

PR #495, opened 2022-07-12
Allow the default pixel_selection method to be overidden 

samn:
in the `MosaicTilerFactory`.

For https://github.com/developmentseed/titiler/discussions/492

The default `PixelSelectionParams` function doesn't change the existing
behavior (it uses `PixelSelectionMethod.first` if no `pixel_selection`
query param is provided), but a custom function can be used to change
the default behavior.


src/titiler/mosaic/tests/test_factory.py
```diff
@@ -113,7 +114,6 @@ def test_MosaicTilerFactory():
             "/mosaic/tiles/7/37/45.npy", params={"url": mosaic_file, "buffer": 10}
         )
         assert response.status_code == 200
-        assert response.status_code == 200
```
vincentsarago, line 116: 🤦 🙏

vincentsarago (approved):

samn:
Thanks @vincentsarago!

Issue #502, opened 2022-08-19
Update tiler with cache code example


catplanck:
In [this example](https://github.com/developmentseed/titiler/blob/master/docs/src/examples/code/tiler_with_cache.md) please add a note about aiocache to use aioredis at version 1.3.1, because the latest aioredis version ([2.0](https://github.com/aio-libs/aioredis-py/releases/tag/v2.0.0)) [doesnt work](https://github.com/aio-libs/aiocache/issues/543) with the current version of aiocache. 

the `@cached()` at step 3 should also be updated to become `@cached(alias="default")`, because if left empty the alias argument in [cached ](https://aiocache.readthedocs.io/en/latest/decorators.html#cached) is defaulted into `None`, thus the default cache config from step 2 wont be used. There should also be a note to explain that with alias argument, `@cached()` will ignore other arguments. for example, you cant add namespace argument inside cached if you use alias (`@cached(namespace="tiles",alias="default")`), it should be declared in the setup cache function from step 2 for it to be used inside the config dict.

please add an example of env file for step 1, ex: `CACHE_ENDPOINT=redis://127.0.0.1:6379/0`, or add a note that says you need to use redis URI in the env file.

please add a note in step 3 that you can use the same step to add cache into mosaic endpoint just by changin the `TilerFactory` function into `MosaicTilerFactory` function (and maybe stac endpoint too, i havent tested it yet).

There should also be an explanation about whether you still have to use `CacheControlMiddleware` together with aiocache or not, they both have Cache in the name after all.

vincentsarago:
Please submit a PR 🙏

PR #503, opened 2022-08-19
Updating tiler with cache example in documentation

catplanck:
In [tile with cache example](https://github.com/developmentseed/titiler/blob/master/docs/src/examples/code/tiler_with_cache.md) documentation

For #502 updating tile with cache example documentation by adding notes about aioredis version, adding env file example, adding alias arguments into `@cached()`, and adding notes about how to add cache into another endpoints.


vincentsarago:
🙏 thanks @catplanck

vincentsarago (approved):

PR #532, opened 2022-11-07
Refactor gis tools

RemiSahl:
## What I am changing
I moved the GIS manipulation logic into a separate module. This allows for importing and using the data manipulation logic in different contexts. Hopefully, I haven't broken anything in the process.

## How I did it
I refactored the factory.py module and moved the data manipulation logic into a separate module (gis_tools.py).
I also modified some of the functions slightly in order to regroup similar code contained in the different subclasses of the TilerFactory (for example, using multi_bands=True or multi_assets=True and calling the same geojson_statistics function).

## How you can test it
By verifying nothing is broken in the various API routes.

## Related Issues
I initially tried to implement a custom /x/y/z route but had to copy the tile fetch logic in my code: https://github.com/developmentseed/titiler/discussions/531. With this PR, I should be able to import the gis_tools.tile function directly and use it in my custom route which should prevent some of the potential maintenance issues in the future.

### DIFF
--- a/src/titiler/core/titiler/core/factory.py
+++ b/src/titiler/core/titiler/core/factory.py
@@ -263,8 +263,7 @@ class TilerFactory(BaseTilerFactory):
             env=Depends(self.environment_dependency),
         ):
             """Return the bounds of the COG."""
-            return gt.call_with_read(
-                func=gt.bounds,
+            return gt.bounds(
                 reader=self.reader,
                 env=env,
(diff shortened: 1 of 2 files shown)


RemiSahl:
I apologize, I initially forgot to run the tests and fixed them in my latest commit.

vincentsarago:
I'm gonna close this, we may retry this later 🙏

PR #539, opened 2022-11-21
Rio tiler v4

vincentsarago:


vincentsarago:
😭 test are failing because pygeos wheels are not available for python 3.10 and 3.11 because the project was merged to shapely...

PR #560, opened 2022-12-08
try proj4leaflet to support projection in `/map` viewer

### LINKED ISSUE #559: Support multiple TMS in `/map` viewer
In titiler.core 0.8 we added a `/map` endpoint which return a simple viewer (where users have to set all parameters as query string in the URL).

template: https://github.com/developmentseed/titiler/blob/master/src/titiler/core/titiler/core/templates/index.html

Right now we only support the `WebMercatorQuad` TMS because we are using maplibre map client. I think it would be possible to use `leaflet` + `proj4leaflet.js` to support all TMS. 

### How to

As shown in https://github.com/developmentseed/timvt/blob/master/demo/leaflet/index_3035.html#L20-L32, we will need to create a CRS object with `proj4leaflet.js` hopefully by using morecantile TMS model

vincentsarago:
closes #559

### DIFF
--- a/src/titiler/core/titiler/core/templates/index.html
+++ b/src/titiler/core/titiler/core/templates/index.html
@@ -4,53 +4,20 @@
         <meta charset='utf-8' />
         <title>TiTiler Map Viewer</title>
         <meta name='viewport' content='initial-scale=1,maximum-scale=1,user-scalable=no' />
-        <script src='https://unpkg.com/maplibre-gl@2.4.0/dist/maplibre-gl.js'></script>
-        <link href='https://unpkg.com/maplibre-gl@2.4.0/dist/maplibre-gl.css' rel='stylesheet' />
-
+        <link rel="stylesheet" href="https://unpkg.com/leaflet@1.9.3/dist/leaflet.css"/>
+        <script src="https://unpkg.com/leaflet@1.9.3/dist/leaflet.js"></script>
+        <script src="https://unpkg.com/proj4@2.3.14/dist/proj4.js"></script>
+        <script src="https://unpkg.com/proj4leaflet@1.0.2/src/proj4leaflet.js"></script>
         <style>
             body { margin:0; padding:0; width:100%; height:100%; background-color: #e5e5e5;}
             #map { position:absolute; top:0; bottom:0; width:100%; }
-            .zoom-info {
-                z-index: 10;
-                position: absolute;
-                bottom: 0;
-                right: 0;
-                padding: 5px;
-                width: auto;
-                height: auto;
-                font-size: 12px;
-                color: #000;
-            }
-            @media(max-width: 767px) {
-              .maplibrectrl-attrib {
-                  font-size: 10px;
-              }
-            }
         </style>
     </head>
     <body>
 
-    <div id='map'>
-      <div class="zoom-info">Zoom: <span id="zoom"></span></div>
-    </div>
+    <div id='map'></div>
 
-    <script>
-
-var map = new maplibregl.Map({
-  container: 'map',
-  style: {
-    version: 8,
-    sources: {},
-    layers: []
-  },
-  center: [0, 0],
-  zoom: 1
-})
-
-map.on('zoom', function (e) {
-  const z = (map.getZoom()).toString().slice(0, 4)
-  document.getElementById('zoom').textContent = z
-})
+    <script type="text/javascript">
 
 const bboxPolygon = (bounds) => {
     return {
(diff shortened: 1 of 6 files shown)


src/titiler/core/titiler/core/templates/index.html
```diff
@@ -69,79 +36,63 @@
...
-      "type": "FeatureCollection",
-      "features": [bboxPolygon(bounds)]
+// tms comes from morecantile
+// https://github.com/developmentseed/morecantile/blob/main/morecantile/models.py
+var crs = new L.Proj.CRS(
+    '{{ tms.crs.srs }}',
+    '{{ tms.crs.to_proj4() }}', {
+    origin: [{{ tms.xy_bbox[0] }}, {{ tms.xy_bbox[3] }}],
+    // bounds: [[{{ tms.xy_bbox[0] }}, {{ tms.xy_bbox[1] }}], [{{ tms.xy_bbox[2] }}, {{ tms.xy_bbox[3] }}]],
+    resolutions: {{ resolutions|safe }},
```
vincentsarago, line 46: ⚠️ need help here ⚠️ 

It seems to work but it needs more testing

vincentsarago: I'm not sure if `resolutions` is https://github.com/developmentseed/morecantile/blob/main/morecantile/models.py#L370-L379

IvanSanchez: So this is a bit tricky.  I'll quote myself from https://gis.stackexchange.com/a/423647/53755 :

> The result of `L.CRS.scale()` has to be multiplied by a CRS-dependant constant (the world bounds according to that CRS) in order to make the numbers meaningful, which IMO leads to confusion.

So you have to be careful about the CRS's bounds, specially if you plan on using TMS-style tiles with an inverted Y-coordinate (a `{-y}` in the URL template). Specifically, see https://stackoverflow.com/a/62625435/4768502

See also https://gis.stackexchange.com/a/310124/53755 - Don't be confused when using Leaflet's functions for projecting/unprojecting coordinates because they don't rely on CRS-relative numbers, but rather screen-relative numbers.

src/titiler/core/titiler/core/templates/index.html
```diff
@@ -69,79 +36,63 @@
...
-    paint: {
-      'line-color': '#3bb2d0',
-      'line-width': 1
-    }
-  })
-  return
-}
+var map = L.map('map', {
+  zoom: 0,
+  center: [0, 0],
```
IvanSanchez, line 53: Don't set zoom & center if you're doing `fitBounds` later. Can lead to undesired glitching/zooming at start.

src/titiler/core/titiler/core/templates/index.html
```diff
@@ -69,79 +36,63 @@
...
-      'line-join': 'round'
-    },
-    paint: {
-      'line-color': '#3bb2d0',
-      'line-width': 1
-    }
-  })
-  return
-}
+var map = L.map('map', {
```
IvanSanchez, line 51: While you're debugging, add a few control points to make sure the CRS mess is working properly. I tend to use stuff like

```
const madrid = L.marker([40, -3]).addTo(map);
```

Also, something very useful is:

```
map.on('click', (ev)=>console.log(ev.latlng));
```

src/titiler/core/titiler/core/templates/index.html
```diff
@@ -69,79 +36,63 @@
...
-    })
-    .catch(err => {
-      console.warn(err)
-    })
-})
+    L.tileLayer(
+      data.tiles[0], {
+        minZoom: data.minzoom,
+        maxZoom: data.maxzoom,
+        bounds: L.latLngBounds([bottom, left], [top, right]),
```
IvanSanchez, line 88: I don't think you want to do this - the `bounds` option on a `L.Tilelayer` is a **constraint**, so the tilelayer will load less tiles. It's not an informative field.

Focus on getting the `bounds` of the Leaflet `CRS` right - tilelayers will work throughout that range, and *only* throughout that range.

src/titiler/core/titiler/core/templates/index.html
vincentsarago: Thanks for you help @IvanSanchez 
going back at this today, and I'm not quite sure I understand all of ☝️ 😭

vincentsarago:
Alright everything seems to work well 🥳 

notes:
- `CanadianNAD83_LCC` does not work but I highly suspect the TMS of not being good 
- `NZTM2000` does not work but I highly suspect the TMS of not being good ame as ☝️ 
- `NZTM2000Quad` also gives weird results 
- `UTM31WGS84Quad` does not I highly suspect the TMS of not being good

PR #572, opened 2023-01-16
WMS prototype working

mindless-bureaucrat:
This PR adds WMS functionality to the base `TilerFactory` at `/wms`

Notes:

- WMS clients send params irregularly as UPPER or lower case. I was unable to find an elegant way for FastAPI to convert all params to the same case, so to mitigate this the lower case (and other WMS param handling) is taken care of in the `process_wms_params` static method. 
- The wms route is split in two parts: GetCapabilities and Get Map
- GetCapabilities will generate a dynamic XML based on the params provided. In addition, multiple COGs can be built in a single XML by using multiple `layers` params. This behavior deviates from the standard `url` param as the primary COG input in the route. 
- GetMap is mostly a copy/paste of the `/tiles` route, but uses `src_dst.part` with bbox rather than `.tile`.
- WMS route is optional, similar to `preview`
- WMS XML templates were added for 1.0.0, 1.1.1, and 1.3.0. . I'm not familiar enough with inner workings of the format, but I suspect someone smarter than I could likely jam it in to one file with some complicated Jinja logic. 
- Additional WMS specific Enums were added


vincentsarago:
Wooot thank for the PR @mindless-bureaucrat 🙏

src/titiler/core/titiler/core/factory.py
```diff
@@ -812,6 +824,299 @@ def wmts(
...
+                out_wms_params['bounds'] = out_wms_params['bbox']
+                    
+            print(out_wms_params)
+            return out_wms_params
+
+        @self.router.get("/wms")
+        def wms(
+            _request: Request,
+            request: WMSRequestType = "GetCapabilities",
+            version: WMSVersion = "1.1.1",
```
vincentsarago, line 936: I've been using `Literal` over enums lately. It gives the same results (for docs and checks)

```suggestion
            version: Literal["1.0.0", "1.1.1", "1.3.0"] = Query("1.1.1", description="WMS Version"),
```

src/titiler/core/titiler/core/factory.py
```diff
@@ -812,6 +824,299 @@ def wmts(
...
+                # `bounds` is compatible with reading from rio.io.reader
+                out_wms_params['bounds'] = out_wms_params['bbox']
+                    
+            print(out_wms_params)
+            return out_wms_params
+
+        @self.router.get("/wms")
+        def wms(
+            _request: Request,
+            request: WMSRequestType = "GetCapabilities",
```
vincentsarago, line 935:
```suggestion
            request_type: Literal["GetCapabilities", "GetMap"] = Query("GetCapabilities", description="WMS request's type."),
```

src/titiler/core/titiler/core/factory.py
```diff
@@ -812,6 +824,299 @@ def wmts(
...
+
+                template = f"wms_{processed_wms_params.get('version')}.xml"
+
+                return templates.TemplateResponse(
+                    template,
+                    {
+                        "request": _request,
+                        "request_url": wms_url,
+                        "formats": [v.value for v in list(WMSMediaType)],
+                        "available_epsgs": [e.value for e in list(EPSGCodes)],
```
vincentsarago, line 1036: in theory all EPSG code found in proj are available not just the ones in morecantile 🤷

src/titiler/core/titiler/core/factory.py
```diff
@@ -812,6 +824,299 @@ def wmts(
...
+                description="COG dataset URL(s) or local path(s) to a COG"
+            ),
+            layer_params=Depends(self.layer_dependency),  # noqa
+            dataset_params=Depends(self.dataset_dependency),  # noqa
+            buffer: Optional[float] = Query(  # noqa
+                None,
+                gt=0,
+                title="Tile buffer.",
+                description="Buffer on each side of the given tile. It must be a multiple of `0.5`. Output **tilesize** will be expanded to `tilesize + 2 * buffer` (e.g 0.5 = 257x257, 1.0 = 258x258).",
+            ),
```
vincentsarago, line 969: buffer is only available for `.tile()` method

src/titiler/core/titiler/core/resources/enums.py
```diff
@@ -72,3 +73,31 @@ class OptionalHeader(str, Enum):
...
     x_assets = "X-Assets"
+    
+class WMSRequestType(str, Enum):
+    GetCapabilities = "GetCapabilities"
+    GetMap = "GetMap"
+
+class WMSVersion(str, Enum):
+    v100 = "1.0.0"
+    v111 = "1.1.1"
+    v130 = "1.3.0"
```
vincentsarago, line 84: we can avoid creating those Enum by using `Literal` in the factory directly

src/titiler/core/titiler/core/factory.py
```diff
@@ -812,6 +824,299 @@ def wmts(
...
+            srs: Union[EPSGCodes, None] = Query(
+                default=None,
+                description="Coordinate system of bbox as an EPSG code",
+            ),
+            # crs: Ignore if we stick with 
+            layers: list[str] = Query(
+                [""], 
+                description="COG dataset URL(s) or local path(s) to a COG"
+            ),
```
vincentsarago, line 961: looking at the OGC specification it seems that layers should be a list of layer delimited by `,`

src/titiler/core/titiler/core/factory.py
```diff
@@ -812,6 +824,299 @@ def wmts(
...
+                description="Width (in pixels) of returned GetMap image."
+            ),
+            height: Union[int, None]  = Query( 
+                default=None,
+                description="Height (in pixels) of returned GetMap image."
+            ),
+            srs: Union[EPSGCodes, None] = Query(
+                default=None,
+                description="Coordinate system of bbox as an EPSG code",
+            ),
```
vincentsarago, line 956: I think we could allow `string` as input, it will be up to the user to gives the right one. rasterio should be able to parse it (either if it's a string or an uri)

src/titiler/core/titiler/core/factory.py
```diff
@@ -812,6 +824,299 @@ def wmts(
...
+            "srs",
+            "width",
+            "height",
+            "layers",
+            "format",
+        )
+        # Ignoring: ['styles', 'dpi', 'map_resolution', 'format_options', 'transparent']
+
+        @staticmethod
+        def process_wms_params(request: Request):
```
vincentsarago, line 852: maybe this whole function could be a Dependency we inject into the wms endpoint

vincentsarago:
```python
@dataclass
class WMSParams(DefaultDependency):

    request: str
    version: str

    layers: Optional[List[str]]
    bbox: Optional[List[float]]
    crs: Optional[CRS]
    height: Optional[int]
    width: Optional[int]
    format: Optional[str]

    def __init__(
        self,
        request: Request,
    ):
        # WMS params use `,` delimited list so we can consider the query-parameter
        # as a dictionary
        req = {
            k.lower(): v
            k, v in for request.query_params
        }

        # Request is mandatory
		if not request_type := req.get("request"):
            raise HTTPException(status_code=400, detail="Missing WMS REQUEST parameter."

        self.request = request_type.lower()

        ogc_spec_version = req.get("version")

        if self.request == "getcapabilities":
            # Required parameters:
            # - SERVICE=WMS
            # - REQUEST=GetCapabilities
            # Optional parameters: VERSION, FORMAT, UPDATESEQUENCE
            if not service := req.get("service"):
                raise HTTPException(status_code=400, detail="Missing WMS SERVICE parameter.")

            if not service.lower() == "wms":
                raise HTTPException(status_code=400, detail=f"Invalid 'SERVICE' parameter: {service}. Only 'wms' is accepted")

            self.version = ogc_spec_version or "1.3.0"

        elif self.request == "getmap":
            # Required parameters:
            # - VERSION
            # - REQUEST=GetMap,
            # - LAYERS
            # - STYLES
            # - CRS
            # - BBOX
            # - WIDTH
            # - HEIGHT
            # - FORMAT
            # Optional parameters: TRANSPARENT, BGCOLOR, EXCEPTIONS, TIME, ELEVATION, ...

            # List of required parameters (styles is excluded)
            req_keys = {"version", "request", "layers", "crs", "bbox", "width", "height", "format"}

            input_keys = set(req.keys())
            intrs = input_keys.intersection(req_keys)
            missing_keys = req_keys.difference(intrs)

            if len(missing_keys) > 0:
                raise HTTPException(status_code=500, detail=f"Missing 'GetMap' parameters: {missing_keys}")

            self.version = ogc_spec_version

            self.layers = list(req["layers"].split(","))
            if not self.layers:
                raise HTTPException(status_code=400, detail=f"Invalid 'LAYERS' parameter: {req['layers']}."

            self.bbox = list(map(float, req["bbox"].split(",")))
            if len(self.bbox) != 4:
                raise HTTPException(
                    status_code=400,
                    detail="Invalid 'BBOX' parameters: {bbox}. Needs 4 coordinates separated by commas",
                )

            self.crs = CRS.from_user_input(req["crs"])
            self.height = int(req["height"])
            self.width = int(req["width"])
            self.format = WMSMediaType(req["format"]).name

            # TODO: check this
            if self.version == "1.3.0":
                # WMS 1.3.0 is lame and flips the coords of EPSG:4326
                if self.crs == CRS.from_epsg(4326):
                    self.bbox = [
                        self.bbox[1],
                        self.bbox[0],
                        self.bbox[3],
                        self.bbox[2],
                    ]
                # Overwrite CRS:84 with EPSG:4326 when specified
                elif self.crs == CRS.from_user_input("CRS:84"):
                    self.crs = CRS.from_epsg(4326)

        else:
            raise HTTPException(status_code=400, detail=f"Invalid 'REQUEST' parameter: {request_type}. Should be one of ['GetCapabilities', 'GetMap']."

        if self.version not in ["1.0.0", "1.1.0", "1.3.0"]:
            raise HTTPException(
                status_code=400,
                detail=f"Invalid 'VERSION' parameter: {version}. Allowed versions include: '1.0.0', '1.1.0', '1.3.0'",
            )
```

src/titiler/core/titiler/core/factory.py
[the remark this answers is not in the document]
mindless-bureaucrat: Don't think I can change this one, as it will break the WMS spec :( . Each WMS client sends a GET request with `request=GetCapabilities` or `request=GetMap` to specify the type of request.

src/titiler/core/titiler/core/factory.py
mindless-bureaucrat: Much cleaner, will remove the enums and change over to literals. 👍

src/titiler/core/titiler/core/factory.py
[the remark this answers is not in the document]
vincentsarago: I've got some change I'll push later today

mindless-bureaucrat: Same comment as above, changing `request` to `request_type` will break WMS clients

vincentsarago: that's why I'm using `alias="request"`

mindless-bureaucrat: ahhhhh! That's a new one for me. ok i like it. Going to do some homework at https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#alias-parameters

src/titiler/core/titiler/core/factory.py
mindless-bureaucrat: Isn't `.part()` imported from here? https://github.com/cogeotiff/rio-tiler/blob/main/rio_tiler/reader.py#L245

src/titiler/core/titiler/core/factory.py
mindless-bureaucrat: Right, I'll change this param to a string

src/titiler/core/titiler/core/factory.py
vincentsarago: 🤦 true, I forgot about this

src/titiler/core/titiler/core/factory.py
mindless-bureaucrat: These are some pro-level edits

mindless-bureaucrat:
> * because the wms do not work exactly like all other endpoint it might make more sense to create a titiler-wms plugin

I initially had WMS as an extension, however the behavior seems similar to the `wmts` endpoint (in that we generate an XML definition and the client then reads from there). I also found maintaining a separate plug-in with the recent changes from TiTiler 0.7  to be cumbersome, however I'll defer to you on this. 

> * what happens when multiple layers are passed to the GetMap request

At this time GetMap will pull the first layer specified (in the string-split list). I know in _real_ WMS there's a way to pull and overlay multiple layers in a single call, however I think that is beyond the scope of just exposing a simple WMS protocol with this addition

src/titiler/core/titiler/core/factory.py
```diff
@@ -824,189 +823,301 @@ def wmts(
...
+        supported_crs: List[str] = ["EPSG:4326"]
+        supported_format: List[str] = [
+            "image/png",
+            "image/jpeg",
+            "image/jpg",
+            "image/webp",
+            "image/jp2",
+            "image/tiff; application=geotiff",
+        ]
+        supported_version: List[str] = ["1.0.0", "1.1.0", "1.3.0"]
```
vincentsarago, line 844: For now this is hardcoded, let's discuss later how users could change ☝️

src/titiler/core/titiler/core/factory.py
```diff
@@ -824,189 +823,301 @@ def wmts(
...
+                        "image/jpeg": {},
+                        "image/jpg": {},
+                        "image/webp": {},
+                        "image/jp2": {},
+                        "image/tiff; application=geotiff": {},
+                    },
+                },
+            },
+            openapi_extra={
+                "parameters": [
```
vincentsarago, line 864: 👇 with openapi_extra we add entries to the openapi.json schema (swagger docs) but this is just for `display` 😄 (the api will accept lower and upper case)

src/titiler/core/titiler/core/factory.py
```diff
@@ -824,189 +823,301 @@ def wmts(
...
-            Return a WMS query for a single COG.
+            """Return a WMS query for a single COG.
 
             GetCapability will generate a WMS XML definition.
 
             GetMap is mostly copied from titiler.core.factory.TilerFactory.part.part
             """
-            processed_wms_params = process_wms_params(_request)
-            print(processed_wms_params)
+            req = {k.lower(): v for k, v in request.query_params.items()}
```
vincentsarago, line 1053: I've removed the middleware function and put everything here directly

src/titiler/core/titiler/core/factory.py
```diff
@@ -1016,44 +1127,123 @@ def wms(
...
+                    raise HTTPException(
+                        status_code=400,
+                        detail=f"Invalid 'LAYERS' parameter: {req['layers']}.",
+                    )
+
+                src_path = layers[0]
                 with rasterio.Env(**env):
                     with self.reader(src_path, **reader_params) as src_dst:
                         image = src_dst.part(
-                            buffer=buffer,
```
vincentsarago: I removed buffer for now but we could add it back

src/titiler/core/titiler/core/factory.py
```diff
@@ -1016,44 +1127,123 @@ def wms(
...
+                    if crs == CRS.from_epsg(4326):
+                        bbox = [
+                            bbox[1],
+                            bbox[0],
+                            bbox[3],
+                            bbox[2],
+                        ]
+                    # Overwrite CRS:84 with EPSG:4326 when specified
+                    elif crs == CRS.from_user_input("CRS:84"):
+                        crs = CRS.from_epsg(4326)
```
vincentsarago, line 1220: I'm not sure I've seen why we need to do this

src/titiler/core/titiler/core/factory.py
```diff
@@ -1016,44 +1127,123 @@ def wms(
...
+                        status_code=400,
+                        detail=f"Invalid 'VERSION' parameter: {version}. Allowed versions include: {supported_version}",
+                    )
+
+                if not set(req.keys()).intersection({"crs", "srs"}):
+                    raise HTTPException(
+                        status_code=500, detail="Missing 'CRS' or 'SRS parameters."
+                    )
+
+                crs = CRS.from_user_input(req.get("crs", req.get("srs")))
```
vincentsarago, line 1199: is `SRS` part of an old specification?

src/titiler/core/titiler/core/factory.py
```diff
@@ -1068,55 +1258,23 @@ def wms(
                 if color_formula:
                     image.apply_color_formula(color_formula)
 
-                if processed_wms_params.get('format'):
-                    imagetype = ImageType(MediaType(processed_wms_params['format'])._name_)
-                else:
-                    imagetype = ImageType.jpeg if image.mask.all() else ImageType.png
```
vincentsarago: the `auto` mode is not really part of the specification so I removed it

src/titiler/core/titiler/core/templates/wms_1.0.0.xml
```diff
@@ -0,0 +1,55 @@
...
+</Service>
+
+<Capability>
+  <Request>
+    <Map>
+      <Format><PNG /><JPEG /></Format>
+      <DCPType>
+        <HTTP>
+          <Get onlineResource="{{ request_url }}" />
+          <Post onlineResource="{{ request_url }}" />
```
vincentsarago, line 17: we don't accept `POST` request so I guess we should remove this

vincentsarago:
with the latest commit I moved the `/wms` endpoint in its own factory and added support for multiple layers

Screen Shot 2023-01-17 at 2 43 52 PM

src/titiler/core/titiler/core/factory.py
```diff
@@ -252,6 +262,7 @@ class TilerFactory(BaseTilerFactory):
     add_preview: bool = True
     add_part: bool = True
     add_viewer: bool = True
+    add_wms: bool = True
```
vincentsarago, line 265:
```suggestion
```

vincentsarago:
@mindless-bureaucrat the more I think about this, the more I feel uncomfortable adding a WMS endpoint in the tiler factories. Here is why:
- I think in the future we will try to support more OGC specification but the WMS is quite old and not really REST friendly (at least less than the WMTS specification)
- Because it uses `Layers` and not `url=` and allow more than one dataset, it will be kinda odd to have the wms endpoint with the `regular` endpoint

I'm not quite sure we want to make a `titiler.wms` module or write just an example. I need to think more about this.

src/titiler/core/titiler/core/templates/wms_1.0.0.xml
```diff
@@ -0,0 +1,55 @@
...
+    <SRS>{{" ".join(available_epsgs)}}</SRS>
+    <LatLonBoundingBox minx="{{service_dict['xmin']}}" miny="{{service_dict['ymin']}}" maxx="{{service_dict['xmax']}}" maxy="{{service_dict['ymax']}}" />
+    <BoundingBox SRS="EPSG:4326"
+                minx="{{service_dict['xmin']}}" miny="{{service_dict['ymin']}}" maxx="{{service_dict['xmax']}}" maxy="{{service_dict['ymax']}}" />
+    {% for layer in layers_dict.keys() %}
+    <Layer queryable="0">
+        <Name>{{layer}}</Name>
+        <Title>{{layer}}</Title>
+        <Abstract>{{layers_dict[layer]['abstract']}}</Abstract>
+        <SRS>{{" ".join(available_epsgs)}}</SRS>
```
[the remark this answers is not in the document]
mindless-bureaucrat: Not sure if it's needed, but this is what I saw in all the auto-generated WMS XMLs from Geoserver and ESRI

mindless-bureaucrat:
> @mindless-bureaucrat the more I think about this, the more I feel uncomfortable adding a WMS endpoint in the tiler factories. Here is why:
> 
> * I think in the future we will try to support more OGC specification but the WMS is quite old and not really REST friendly (at least less than the WMTS specification)
> * Because it uses `Layers` and not `url=` and allow more than one dataset, it will be kinda odd to have the wms endpoint with the `regular` endpoint
> 
> I'm not quite sure we want to make a `titiler.wms` module or write just an example. I need to think more about this.

100% agree that WMS has seen better days. Unfortunately I (and likely others) are working in an environment with legacy web clients. This plugin/extension allows us to support modern clients (tile/rest based) along with dinosaurs on the same data definitions without having to run Geoserver (or similar WMS server) in addition to our beautiful TiTiler and FastAPI endpoints :) .

vincentsarago:
> Happy to have this functionality as a separate factory! 

👍 ok then 

### ToDo
- double check that we are OGC compliant
- add tests
- add WMSFactory for titiler.mosaic 😄

vincentsarago:
so in parallel I've started https://github.com/developmentseed/titiler/pull/575, if it goes well I'll add a wmsExtension to the package with most of the code we added here.

then we could use it as 

```python
from fastapi import FastAPI
from titiler.extensions import wmsExtension

app = FastAPI()

tiler = TilerFactory(
    extensions=[
        wmsExtension()
    ]
)

app.include_router(tiler.router)
```

vincentsarago:
I've moved all the WMS related code to an extension.

TODO:
- add getTile tests
- add Extension documentation

vincentsarago:
@mindless-bureaucrat would you be able to review the code before we merge? 🙏 

I'll take care of the documentation in another PR

src/titiler/extensions/titiler/extensions/wms.py
```diff
@@ -0,0 +1,535 @@
...
+                crs = CRS.from_user_input(req.get("crs", req.get("srs")))
+
+                bbox = list(map(float, req["bbox"].split(",")))
+                if len(bbox) != 4:
+                    raise HTTPException(
+                        status_code=400,
+                        detail=f"Invalid 'BBOX' parameters: {req['bbox']}. Needs 4 coordinates separated by commas",
+                    )
+
+                # TODO: check this
```
vincentsarago, line 451:
```suggestion
```

src/titiler/extensions/tests/test_wms.py
```diff
@@ -0,0 +1,373 @@
...
+        assert meta["count"] == 2
+
+        response = client.get(
+            "/wms",
+            params={
+                "VERSION": "1.3.0",
+                "REQUEST": "GetMap",
+                "LAYERS": cog,
+                "BBOX": "373185.0,8019284.949381611,639014.9492102272,8286015.0",
+                "SRS": "EPSG:32621",
```
mindless-bureaucrat, line 228: WMS 1.3.0 uses `crs` instead of `srs`. Annoying but small change. Geoserver will accept either just for the sake of flexibility. I suggest we stick with interchangeable SRS/CRS, unless you want to go by the OGC spec. For reference: https://docs.geoserver.org/stable/en/user/services/wms/basics.html

src/titiler/extensions/tests/test_wms.py
```diff
@@ -0,0 +1,373 @@
...
+                "CRS": "EPSG:32621",
+                "WIDTH": 334,
+                "HEIGHT": 333,
+                "FORMAT": "image/png",
+            },
+        )
+        assert response.status_code == 400
+        assert "Invalid 'BBOX' parameters" in response.text
+
+        # 1.3.0 needs inverted X,Y coordinates for EPSG:4326
```
mindless-bureaucrat, line 305: Awesome checking for the messed up coordinate swapping when switching to 1.3.0 and 4326

src/titiler/extensions/titiler/extensions/wms.py
```diff
@@ -0,0 +1,534 @@
...
+                        status_code=400,
+                        detail=f"Invalid 'VERSION' parameter: {version}. Allowed versions include: {self.supported_version}",
+                    )
+
+                if not set(req.keys()).intersection({"crs", "srs"}):
+                    raise HTTPException(
+                        status_code=400, detail="Missing 'CRS' or 'SRS parameters."
+                    )
+
+                crs = CRS.from_user_input(req.get("crs", req.get("srs")))
```
mindless-bureaucrat, line 442: Should we raise HTTPException if crs ends up being empty?

src/titiler/extensions/titiler/extensions/wms.py
```diff
@@ -0,0 +1,534 @@
...
+                                bbox,
+                                width=width,
+                                height=height,
+                                dst_crs=crs,
+                                bounds_crs=crs,
+                                **layer_params,
+                                **dataset_params,
+                            )
+
+                image, assets_used = mosaic_reader(
```
mindless-bureaucrat, line 503: In theory we should only be passing a single layer in the `layers` qs param, although I supposed this could work for a very simple mosaic method, absent MosaicJson. Is this the intended purpose rather than just getting the image out with .part() directly? How would this work with the MosaicTilerFactory if a user specified a custom backend?

src/titiler/extensions/tests/test_wms.py
```diff
@@ -0,0 +1,373 @@
...
+        )
+        assert response.status_code == 200
+        meta = parse_img(response.content)
+        assert meta["crs"] == CRS.from_epsg(4326)
+        assert meta["bounds"] == [
+            -61.28762442711404,
+            72.22979795551834,
+            -52.301598718454485,
+            74.66298001264106,
+        ]
```
mindless-bureaucrat, line 373: Should we implement the mosaic method in this PR as well, or wait until a future version? I have a working prototype in my local rats nest, happy to wait or push to a later date :)

mindless-bureaucrat (commented):
You are a machine! Great looking tests and cleanup. Have a couple comments regarding using the mosaic method rather than .part, and a few WMS spec related things.

src/titiler/extensions/tests/test_wms.py
vincentsarago: I didn't know if this was version related. 👍 for supporting both whatever the version is

src/titiler/extensions/titiler/extensions/wms.py
vincentsarago: we could 👍

src/titiler/extensions/titiler/extensions/wms.py
vincentsarago: so here we use simple mosaic, meaning you just need to pass multiple file name, there is no idea of mosaic json or mosaic backend. 

> How would this work with the MosaicTilerFactory if a user specified a custom backend?

It won't work for mosaic backend because they don't implement `.part()` method (I might add it in titiler-pgstac)

src/titiler/extensions/tests/test_wms.py
vincentsarago: because mosaic backend do not implement `part()` method, I'm not planing to add it

vincentsarago:
@mindless-bureaucrat I'm going to merge this PR, please if you have more comments feel free to raise them before we make the next release 🙏

PR #575, opened 2023-01-18
add titiler extensions

vincentsarago:
This PR add a new titiler package `extensions` which goal is to ease the customization of Tiler Factories (read: add endpoints). 

This will be pretty useful for #572

### TL&DR

```python
from fastapi import FastAPI
from titiler.extensions import cogValidateExtension

# Create a FastAPI application
app = FastAPI(description="A lightweight Cloud Optimized GeoTIFF tile server")

# Create a set of endpoints using TiTiler TilerFactory
tiler = TilerFactory(
    extensions=[
        cogValidateExtension()  # the cogeoExtension will add a rio-cogeo /validate endpoint
    ]
)

# Register endpoints to the application
app.include_router(tiler.router)
```

in ☝️, we add the `cogValidateExtension` which will then add a `/validate` endpoint. The important thing are:
- if a tiler has a prefix, then the extension endpoint will also have the prefix
- extension endpoints can use the same dependencies as set in the TilerFactory (e.g path_dependency)

see https://github.com/developmentseed/titiler/blob/21a196fb80ab46cda1d468575a94b43629ae0085/src/titiler/extensions/README.md for more info


geospatial-jeff (approved):
Really simple, I like it!

PR #578, opened 2023-01-26
add Strict mode for mosaic zoom level

vincentsarago:
ref https://github.com/developmentseed/titiler/discussions/577

### DIFF
--- a/src/titiler/mosaic/titiler/mosaic/factory.py
+++ b/src/titiler/mosaic/titiler/mosaic/factory.py
@@ -26,7 +26,7 @@ from titiler.core.resources.responses import GeoJSONResponse, JSONResponse, XMLR
 from titiler.mosaic.models.responses import Point
 from titiler.mosaic.resources.enums import PixelSelectionMethod
 
-from fastapi import Depends, Path, Query
+from fastapi import Depends, HTTPException, Path, Query
 
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, Response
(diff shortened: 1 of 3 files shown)


src/titiler/mosaic/titiler/mosaic/factory.py
```diff
@@ -278,13 +271,22 @@ def tile(
             """Create map tile from a COG."""
             threads = int(os.getenv("MOSAIC_CONCURRENCY", MAX_THREADS))
 
+            strict_zoom = str(os.getenv("MOSAIC_STRICT_ZOOM", False)).lower() == "true"
```
vincentsarago, line 274: By default we keep the `old` behaviour

vincentsarago: maybe we should allow other value than `TRUE/True/true` like `YES/Yes/yes`

mradamcox: I can't build and test this at the moment, but it looks perfectly good from here :). `TRUE/True/true` seems like plenty to me, though I'm not sure what the conventions are for other settings in Titiler/rio-tiler/GDAL, etc. (Elsewhere I have seen
```
import ast
strict_zoom = ast.literal_eval(os.getenv("MOSAIC_STRICT_ZOOM", False))
# strict_zoom will be Bool()
```
used to enforce env variable values.)

geospatial-jeff (approved):
LGTM

mradamcox (approved):

src/titiler/mosaic/titiler/mosaic/factory.py
vincentsarago: > `ast.literal_eval(os.getenv("MOSAIC_STRICT_ZOOM", "False"))`

if `MOSAIC_STRICT_ZOOM` is set to `TRUE` or `FALSE` this will error -> ValueError: malformed node or string: <ast.Name object at 0x10ea61850>

mradamcox: Yeah, that approach would be more strict, only working with `True` or `False`. No matter to me, just wanted to suggest it.

PR #583, opened 2023-02-08
[TO BE DISCUSSED] sketch TMS scheme support

vincentsarago:
ref #https://github.com/developmentseed/titiler/discussions/582

I'm not really familiar with the TMS scheme so if anyone can confirm that this is doing what is expected I would be grateful

### DIFF
--- a/src/titiler/core/titiler/core/dependencies.py
+++ b/src/titiler/core/titiler/core/dependencies.py
@@ -3,10 +3,11 @@
 import json
 from dataclasses import dataclass
 from enum import Enum
-from typing import Dict, List, Optional, Sequence, Tuple, Union
+from typing import Dict, List, Literal, Optional, Sequence, Tuple, Union
 
 import numpy
-from fastapi import HTTPException, Query
+from fastapi import HTTPException, Path, Query
+from morecantile import Tile
 from rasterio.enums import Resampling
 from rio_tiler.colormap import cmap, parse_color
 from rio_tiler.errors import MissingAssets, MissingBands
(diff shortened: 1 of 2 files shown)


src/titiler/core/titiler/core/dependencies.py
```diff
@@ -445,3 +446,18 @@ def __post_init__(self):
...
+    z: int = Path(..., ge=0, le=30, description="Tiles's zoom level"),
+    x: int = Path(..., description="Tiles's column"),
+    y: int = Path(..., description="Tiles's row"),
+    scheme: Optional[Literal[("xyz", "tms")]] = Query(
+        None, description="Tile Scheme (default to XYZ)."
+    ),
+) -> Tile:
+    """Tile parameters."""
+    if scheme == "tms":
+        y = 2**z - y - 1
```
vincentsarago, line 461: 👀 from https://gist.github.com/tmcw/4954720

guillemc23: To my understanding it goes that way, since `y_TMS = 2^z - y_XYZ -1` but also `y_XYZ = 2^z - y_TMS -1`. From the Titiler docs, isn't the default TMS? 
image

vincentsarago: > From the Titiler docs, isn't the default scheme TMS?

there is a difference between `OGC TMS -> TileMatrixSet` and `Osgeo TMS -> Tile Map Service` which is what the scheme is. 

in TiTiler we use the acronyme TMS as the OGC one `TileMatrixSet`

guillemc23: Okay, alright, now I get it 👌

PR #586, opened 2023-02-11
docs: Fix recommended CPL_VSIL_CURL_CACHE_SIZE

neilsh:
Value was equivalent to 20MB instead of desired 200MB


vincentsarago:
🙏 thanks

PR #602, opened 2023-03-09
add left and right sources to visualize dateline crossing dataset

vincentsarago:
This PR `fixes` the cog/stac viewer to handle dataset crossing the dateline by using 2 different Layers on each side of the line.

Screenshot 2023-03-09 at 2 52 52 AM


emmanuelmathot:
ok I give a look

PR #603, opened 2023-03-09
Conditional Chart install test

emmanuelmathot:
The chart install step should not execute if no change has been made in the chart. The condition was not present and the action tried to install the chart in a not initialized cluster. This PR fixes the issue.


vincentsarago:
🙏 Thank you

PR #608, opened 2023-03-16
%matplotlib inline

samsammurphy:
in response to this warning..

`%pylab is deprecated, use %matplotlib inline and import the required libraries. Populating the interactive namespace from numpy and matplotlib`


vincentsarago (approved):

vincentsarago:
Thanks @samsammurphy 🙏

PR #612, opened 2023-03-21
sketch use of Annotated types

vincentsarago:
Possible fix for using Optional Path parameter when multiple routes are defined 

ref https://github.com/tiangolo/fastapi/discussions/9278


vincentsarago:
While the tests pass, I don't think the solution is 👍 

Removing the use of `Query/Path` and use simple Annotated type means that we are looking OpenAPI description/title.

I've opened https://github.com/tiangolo/fastapi/pull/9670 to see if we can do something about it

vincentsarago:
I'm going to move forward with this PR because:
- using `Annotated` seems to be the new default way of doing things in pydantic/fastapi
- We don't want to be blocked at fastapi 0.95 
- We want to be able to use fastapi>0.100 with pydantic 2.0

PR #619, opened 2023-03-29
Allow a default rescale parameter to be set via a dependency

samn:
## What I am changing
- Add a `rescale_dependency` field to `BaseTilerFactory` & `MosaicTilerFactory` (defaulting to the existing `RescalingParams`) so users can override the logic for supplying the `rescale` value. This lets us define a default rescale value from a separate config without needing to include it in the query params for every request.

## How I did it
- Added a dependency to get the `rescale` value, rather than using a `Query` directly that can't be overridden. 

## How you can test it
- Via the integration tests, or by using the `rescale_dependency` in an existing application.


vincentsarago:
🙏 nice addition @samn 

would you be able to add something in the changelog (under a `##Next` item) 🙏

vincentsarago:
thank you @samn 🙏

PR #632, opened 2023-04-14
add coord-crs options

### LINKED ISSUE #631: Point endpoint only accepts coordinates in EPSG:4326
#### Problem description

using the point value endpoint, I noticed that there is no way to specify the CRS of the lon/lat of the input leading to issues when working with non-earth CRSs ([example](https://astrogeology.usgs.gov/titiler/titiler/titiler/cog/point/0,0?url=https%3A%2F%2Fmurray-lab.caltech.edu%2Ftemp%2Fannex%2Fmola_color_1km_tms.tif&bidx=1&expression=b1%2Fb2&unscale=false&resampling=nearest)). 

the rio_tiler point method does accept a `coord_crs` parameter that could be used to address this issue.
I'd hope it would be possible to update titiler to use the geodetic_crs for the COG or to allow the user to specify the CRS as an optional parameter. 

That could also be useful if the user wanted to provide coordinates in a different projection/unit but could be an over compilation. Alternatively another endpoint could be added to accept a query point in native coordinates (with no attempts to reproject)

#### Expected Output

correct point values for various CRSs

vincentsarago:
closes #631 

This PR adds a `coord-crs=` option to `/point`, `/crop`, `/feature` and `/statistics[POST]` method to allow user to specify the CRS for the input coordinates.

Notes: Sadly it's not as simple for the mosaic endpoint (see https://github.com/developmentseed/cogeo-mosaic/issues/209)

@AndrewAnnex could you double check that's it works for you 🙏

### DIFF
--- a/src/titiler/core/titiler/core/factory.py
+++ b/src/titiler/core/titiler/core/factory.py
@@ -14,6 +14,8 @@ from geojson_pydantic.geometries import Polygon
 from morecantile import TileMatrixSet
 from morecantile import tms as morecantile_tms
 from morecantile.defaults import TileMatrixSets
+from rasterio.crs import CRS
+from rio_tiler.constants import WGS84_CRS
 from rio_tiler.io import BaseReader, MultiBandReader, MultiBaseReader, Reader
 from rio_tiler.models import BandStatistics, Bounds, Info
 from rio_tiler.types import ColorMapType
@@ -35,6 +37,7 @@ from titiler.core.dependencies import (
     BandsParams,
     BidxExprParams,
     ColorMapParams,
+    CRSParams,
     DatasetParams,
     DatasetPathParams,
     DefaultDependency,
@@ -367,6 +370,7 @@ class TilerFactory(BaseTilerFactory):
             with rasterio.Env(**env):
                 with self.reader(src_path, **reader_params) as src_dst:
                     return Feature(
+                        type="Feature",
                         geometry=Polygon.from_bounds(*src_dst.geographic_bounds),
                         properties=src_dst.info(),
                     )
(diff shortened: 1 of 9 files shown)


src/titiler/core/titiler/core/factory.py
```diff
@@ -367,6 +370,7 @@ def info_geojson(
             with rasterio.Env(**env):
                 with self.reader(src_path, **reader_params) as src_dst:
                     return Feature(
+                        type="Feature",
```
vincentsarago, line 373: This change is unrelated to the PR goal but it will prevent breaking change with the next version of geojson-pydantic

src/titiler/core/titiler/core/factory.py
```diff
@@ -439,13 +444,14 @@ def geojson_statistics(
...
             if isinstance(fc, Feature):
-                fc = FeatureCollection(features=[geojson])
+                fc = FeatureCollection(type="FeatureCollection", features=[geojson])
 
             with rasterio.Env(**env):
                 with self.reader(src_path, **reader_params) as src_dst:
                     for feature in fc:
                         data = src_dst.feature(
                             feature.dict(exclude_none=True),
+                            shape_crs=coord_crs or WGS84_CRS,
```
vincentsarago, line 454: we will always default to `WGS84_CRS` (we can't pass `None` because rio-tiler won't like it)

AndrewAnnex: small question possibly because I am naive about fastapi, but would it be possible/better to set the coord_crs default value to WGS84_CRS rather than using ors inside the function call?

AndrewAnnex:
looks good to me @vincentsarago, I just have a small question about using the or statements vs setting a default value using fastapi's mechanisms via a comment I posted.

src/titiler/core/titiler/core/factory.py
vincentsarago: The issue is that if you set a default in the `Query` parameter it will appear in the OpenAPI documentation and be added in the URLs by default which I find annoying (we can see this with the `return_mask` option). 

Also in the future we will want to only rely on rio-tiler's default (right now it won't like if you pass `None`).

vincentsarago:
Tests are failing due to some url encoding issue, I'll have a look later

src/titiler/core/titiler/core/factory.py
AndrewAnnex: okay wasn't aware of that, sounds good to me then!

PR #657, opened 2023-06-08
Added securityContext for k8s

holgerbach:
SecurityContext needed for hardening, which you can define in your values.yaml as follows:

```
securityContext:
  capabilities:
    drop:
      - ALL
  readOnlyRootFilesystem: true
  allowPrivilegeEscalation: false
  runAsNonRoot: true
  runAsUser: 1000

podSecurityContext:
  readOnlyRootFilesystem: true
  runAsNonRoot: true
  runAsUser: 1000 
```

### DIFF
--- a/deployment/k8s/charts/Chart.yaml
+++ b/deployment/k8s/charts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 0.11.7
 description: A dynamic Web Map tile server
 name: titiler
-version: 1.1.1
+version: 1.1.2
 icon: https://raw.githubusercontent.com/developmentseed/titiler/main/docs/logos/TiTiler_logo_small.png
 maintainers:
   - name: emmanuelmathot  # Emmanuel Mathot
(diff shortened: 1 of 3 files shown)


vincentsarago:
@ividito @ranchodeluxe would you be able to review this PR? 🙏

ranchodeluxe (changes requested):
All this looks good @holgerbach thanks for adding those changes 💯  A couple things:

1. can you add `securityContext: {}` and `podSecurityContext: {}` as defaults to the `values.yml` please so this doesn't create empty blocks 
2. And please bump the patch number of the chart version https://github.com/developmentseed/titiler/blob/main/deployment/k8s/charts/Chart.yaml#L5 (similar to how [this PR](https://github.com/developmentseed/titiler/pull/660/files) did)

ranchodeluxe:
@holgerbach: I see your [previous commit added an empty](https://github.com/developmentseed/titiler/pull/657/commits/1f04d19d66c66ad574dcee266b461f0e4b15a457) `{}` but then you changed your mind.

I don't think most folks will want to use `securityContext:` by default. Do you mind changing it back to `{}` as the default?

ranchodeluxe (approved):
@holgerbach: LGTM now though it seems there's a merge conflict to be resolved around Chart.yml version. Once that's fixed go ahead and merge please. 

Note for future us: we'll probably want CI to automatically bump Chart versions for us in the future

PR #684, opened 2023-08-11
Corse allow methods in env options. #683

ubi15:
As for in discussion #683 

The default settings don't allow to use the `statistics` with geojson `POST` endpoint from a web-app, as only the `GET` method is allowed.

This pull request includes a three-lines change to allow users to allow also the `POST` method by setting `TITILER_API_CORS_ALLOW_METHODS='GET,POST'`.

I know about the ongoing discussion on whether or not to include this middleware in the default application. I am ok with my contribution being rejected if it is not in line with your ideas.

However, I would have saved a day of development if this option were available and documented and it might be help other people not expert in HTTP request to quickly test and deploy their own `titiler` backend.


vincentsarago:
🙏 thanks @ubi15

Issue #685, opened 2023-08-11
Load first three image bands as RGB


tayden:
#### Problem description

Currently, the TiTiler viewer loads the first COG band for the red channel, the second band for the green channel, and the final (e.g. 4th band in a 4 band image) as the blue channel. I would like to see this changed to just load the first three bands as the RGB channels, instead of the unusual behaviour for loading the blue channel that currently exists.

![image](https://github.com/developmentseed/titiler/assets/6541671/7ded6e31-932c-47f3-8837-402d1cefa924)

For imagery that, e.g. include an NIR channel, the channel order is typically R-G-B-NIR (or B-G-R-NIR, sometimes). In my opinion, the bands I would want to load are nearly never the first, second, and fourth bands for visualization purposes, and loading the first, second, and third bands is a much better default since it would usually load a true-color image.

#### Expected Output
By default:

![image](https://github.com/developmentseed/titiler/assets/6541671/e7c5e2b7-09d6-4764-9754-8f4d117bfedd)

I'm happy to submit a PR for this, if it helps get it done.

vincentsarago:
@tayden PR is more than welcomed 🙏

vincentsarago:
done in #688

PR #688, opened 2023-08-21
Simplify band loading

tayden:
## What I am changing

- Simplify and improve default band selection in RGB viewer for both COG and STAC images

## How I did it

- Removed the special band selection logic that applied to the blue band but not the others.
- Simplified the loading logic so the first three image bands are loaded into the R+G+B channels, respectively, in the viewer.
- Tested by running the app and manually checking things worked as expected

## How you can test it

- Run the server, load a COG or STAC item

## Related Issues

- #685

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 - replace mapbox-gl by maplibre
 - replace Stamen by OpenStreetMap tiles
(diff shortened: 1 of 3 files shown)


vincentsarago:
🙏 thanks @tayden

PR #697, opened 2023-09-13
basic WMS Getfeatureinfo implementation

benjaminleighton:
## What I am changing

I'm adding a basic GetFeatureInfo implementation to the Titiler WMS extension. This provides compatibility so that various WMS clients can automatically provide a users with queried point values. 

## How I did it

GetFeatureInfo reuses most of the logic from GetMap and queries a point within the returned image tile. 

## How you can test it

You can test this using the cog/wms endpoint in a WMS client like QGIS or nationalmap.gov.au and using the feature info capabilities to query a point. For some reason in QGIS some datasets like https://oin-hotosm.s3.amazonaws.com/56f9b5a963ebf4bc00074e70/0/56f9c2d42b67227a79b4faec.tif require invert axis orientation to be selected.

### DIFF
--- a/src/titiler/extensions/titiler/extensions/wms.py
+++ b/src/titiler/extensions/titiler/extensions/wms.py
@@ -145,7 +145,8 @@ class wmsExtension(FactoryExtension):
                             "title": "Output format of service metadata/map",
                             "type": "string",
                             "enum": [
-                                "text/html" "application/xml",
+                                "text/html",
+                                "application/xml",
                                 "image/png",
                                 "image/jpeg",
                                 "image/jpg",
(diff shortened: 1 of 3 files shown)


src/titiler/extensions/titiler/extensions/wms.py
```diff
@@ -530,8 +535,48 @@ def _reader(src_path: str):
...
+                    "j",
+                }
+                version, image, format, transparent = get_map_data(req, req_keys)
+                if version == "1.3.0":
+                    i = int(req["i"])
+                    j = int(req["j"])
+
+                    num_bands = image.data.shape[
+                        0
+                    ]  # assuming the image data shape is (num_bands, height, width)
```
vincentsarago, line 567: here we coud directly do 
```
num_bands = image.count
```

https://github.com/cogeotiff/rio-tiler/blob/main/rio_tiler/models.py#L529-L532

vincentsarago:
done in #698

Issue #703, opened 2023-09-26
Error at startup when using pydantic 2.4.0


tronto20:
#### Problem description

[pydantic v2.4.0](https://github.com/pydantic/pydantic/releases/tag/v2.4.0) released recently, and it seems to be causing an error.

error log:
```
...
[2023-09-26 13:59:06 +0000] [1] [INFO] Using worker: uvicorn.workers.UvicornWorker
[2023-09-26 13:59:06 +0000] [7] [INFO] Booting worker with pid: 7
[2023-09-26 13:59:08 +0000] [7] [ERROR] Exception in worker process
Traceback (most recent call last):
  File "/opt/bitnami/python/lib/python3.11/site-packages/gunicorn/arbiter.py", line 589, in spawn_worker
    worker.init_process()
  File "/opt/bitnami/python/lib/python3.11/site-packages/uvicorn/workers.py", line 66, in init_process
    super(UvicornWorker, self).init_process()
  File "/opt/bitnami/python/lib/python3.11/site-packages/gunicorn/workers/base.py", line 134, in init_process
    self.load_wsgi()
  File "/opt/bitnami/python/lib/python3.11/site-packages/gunicorn/workers/base.py", line 146, in load_wsgi
    self.wsgi = self.app.wsgi()
                ^^^^^^^^^^^^^^^
  File "/opt/bitnami/python/lib/python3.11/site-packages/gunicorn/app/base.py", line 67, in wsgi
    self.callable = self.load()
                    ^^^^^^^^^^^
  File "/opt/bitnami/python/lib/python3.11/site-packages/gunicorn/app/wsgiapp.py", line 58, in load
    return self.load_wsgiapp()
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/bitnami/python/lib/python3.11/site-packages/gunicorn/app/wsgiapp.py", line 48, in load_wsgiapp
    return util.import_app(self.app_uri)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/bitnami/python/lib/python3.11/site-packages/gunicorn/util.py", line 359, in import_app
    mod = importlib.import_module(module)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/bitnami/python/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
[...]

PR #704, opened 2023-09-27
rename /crop endpoints to /bbox or /feature

### LINKED ISSUE #702: rename `/crop` -> `/bbox` and `/feature` ?
In the tiler factory we create a set of endpoint with `/crop` prefix to handle both `bbox` (GET) and `feature` (POST) image queries.

https://github.com/developmentseed/titiler/blob/108283f7d4366394e5d32831055d2d9982de4e41/src/titiler/core/titiler/core/factory.py#L930-L938

https://github.com/developmentseed/titiler/blob/108283f7d4366394e5d32831055d2d9982de4e41/src/titiler/core/titiler/core/factory.py#L993-L1005

I'm not a big `fan` of the `/crop` name and I think it could make sense to change the endpoint prefix to something with more meanings

vincentsarago:
closes #702 

This PR does:

- rename `/crop` endpoints to both `/bbox` and `/feature` endpoint **breaking**
- replace `img_dependency` with `img_preview_dependency` and `img_part_dependency` **breaking**
- remove `1024` default for `max_size` when using `bbox` and `feature` endpoints


src/titiler/core/pyproject.toml
```diff
@@ -33,7 +33,8 @@ dependencies = [
     "geojson-pydantic>=1.0,<2.0",
     "jinja2>=2.11.2,<4.0.0",
     "numpy",
-    "pydantic~=2.0",
+    # https://github.com/pydantic/pydantic/pull/7653
+    "pydantic==2.3",
```
vincentsarago, line 37: ref #703 

This will be fixed with pydantic 2.4.2

src/titiler/core/titiler/core/dependencies.py
```diff
@@ -320,6 +320,20 @@ def __post_init__(self):
             self.max_size = None
 
 
+@dataclass
+class PartFeatureParams(DefaultDependency):
```
vincentsarago, line 324: New Dependency for `/bbox`, `/feature` and `/statistics [POST]` endpoint where we don't set `max_size` default

geospatial-jeff: `PartFeatureParams` is a confusing name but I can't really thing of anything better myself 🤷

geospatial-jeff (commented):
LGTM

PR #707, opened 2023-10-03
Allow a default color_formula to be defined on the factory

samn:
instead of needing it to be passed through query params.

## What I am changing
- Add a `color_formula_dependency` field to `BaseTilerFactory` & `MosaicTilerFactory` (defaulting to the existing `ColorFormulaParams`) so users can override the logic for supplying a color formula. This lets us define a default color formula from a separate config without needing to include it in the query params for every request.

## How I did it
- Added a dependency to get the `color_formula` value, rather than using a `Query` directly that can't be overridden.

## How you can test it
- Via the integration tests, or by using the `color_formula_dependency` in an existing application.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Release Notes
 
+## Next Release
+
+* Allow a default `color_formula` parameter to be set via a dependency (author @samn, https://github.com/developmentseed/titiler/pull/707)
+
+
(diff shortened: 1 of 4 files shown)


vincentsarago (approved):

vincentsarago:
Thank you @samn 🙏

PR #713, opened 2023-10-17
Proposal: Change 404 to 204 NO CONTENT when no asset is found (mosaicjson tiling)

simouel:
Currently, when we mosaic some distant rasters in a large area, leaflet raises a 404 error in the console for every missing tile. Changing from 404 to 204 seems to fix the problem

## Available PR templates

- _Alternatively delete and start empty_


vincentsarago:
sounds good to me 🙏 

thanks @simouel

vincentsarago (approved):

PR #725, opened 2023-11-06
Aws private api deployment

Jeandsmith:
## What I am changing
Added an example for a deployment of a AWS private rest-api based on public-sector use developers. 

## How you can test it
No tests, simply deploy after setting the `vpc_endpoint_id` config value.


vincentsarago:
that's interesting! from my understanding, the lambda is not directly within the VPC but only the API gateway?

Jeandsmith:
The vpc endpoint security group can be configured to allow inbound requests from the Lambda function

deployment/aws/cdk/read.14868.1.lock
```diff
@@ -0,0 +1 @@
+14868
```
vincentsarago, line 1: I'm not sure what is this file and why it has to be included

deployment/aws/cdk/config.py
```diff
@@ -85,4 +85,7 @@ class StackSettings(BaseSettings):
     # Default: - No specific limit - account limit.
     max_concurrent: Optional[int] = None
 
+    # The VPC Endpoint ID
+    vpc_endpoint_id: str = "{{VPCE_ID}}"
```
vincentsarago, line 89: This should be optional (default to None), we can then add a check within the `TitilerPrivateApiStack` construct

deployment/aws/cdk/read.14868.1.lock
Jeandsmith: Removed. Not sure what was it either.

deployment/aws/cdk/config.py
Jeandsmith: Changed the config to optional and adapted the construct to it. Should be good.

PR #735, opened 2023-12-05
Add global access token setting and validation

DeflateAwning:
This PR is a feature addition, which allows the addition of a global access token to all endpoints.

This feature is important because it allows implementing a basic security, in cases where the endpoint is routed through a machine-to-machine connection. For example, a Ruby API handles all incoming requests and performs the user permissions validation, and then proxies the requests onto a TiTiler server; no requests from outside the Ruby API should be allowed to access the TiTiler endpoints directly.

Please advise if any changed are required for acceptance of this small feature addition!

### DIFF
--- a/src/titiler/application/titiler/application/main.py
+++ b/src/titiler/application/titiler/application/main.py
@@ -3,7 +3,7 @@
 import logging
 
 import jinja2
-from fastapi import FastAPI, Depends, HTTPException, Security
+from fastapi import Depends, FastAPI, HTTPException, Security
 from fastapi.security.api_key import APIKeyQuery
 from rio_tiler.io import STACReader
 from starlette.middleware.cors import CORSMiddleware
@@ -51,13 +51,15 @@ api_settings = ApiSettings()
 ###############################################################################
 # Setup a global API access key, if configured
 api_key_query = APIKeyQuery(name="access_token", auto_error=False)
+
+
 def validate_access_token(access_token: str = Security(api_key_query)):
     """Validates API key access token, set as the `api_settings.global_access_token` value.
     Returns True if no access token is required, or if the access token is valid.
     Raises an HTTPException (403) if the access token is required but invalid/missing."""
     if api_settings.global_access_token is None:
         return True
-    
+
     if not access_token:
         raise HTTPException(status_code=403, detail="Missing `access_token`")
 
@@ -66,6 +68,8 @@ def validate_access_token(access_token: str = Security(api_key_query)):
         raise HTTPException(status_code=403, detail="Invalid `access_token`")
 
     return True
+
+
 ###############################################################################
 
 app = FastAPI(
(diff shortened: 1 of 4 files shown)


vincentsarago:
Dear @DeflateAwning 

We appreciate the contribution. 

Adding Auth on top of titiler application should be done at user level. We see `titiler.application` module as a demo not as something that should get used in production and specifically if users required auth. 

`Global access token` is pretty weak auth and I don't think it provide real security (IMO) which is why in the documentation we added https://developmentseed.org/titiler/examples/code/tiler_with_auth/ to show how users can implement auth on their own application.

vincentsarago:
@DeflateAwning can you resolve the issue to make sure the code works with python 3.8 🙏

DeflateAwning:
Done! Forgot about that old Python3.8/3.9 type hinting syntax

vincentsarago:
@DeflateAwning can you run pre-commit 🙏

DeflateAwning:
Hmm, I ran all the content in [CONTRIBUTING.md](https://github.com/developmentseed/titiler/blob/main/CONTRIBUTING.md) before the Docs section, and then ran `pre-commit`, but there are no changes detected (on the latest version of this code). Is there something I'm missing? Was in Python3.10 venv.

src/titiler/application/titiler/application/main.py
```diff
@@ -47,6 +48,30 @@
...
+    Raises an HTTPException (403) if the access token is required but invalid/missing."""
+    if api_settings.global_access_token is None:
+        return True
+
+    if not access_token:
+        raise HTTPException(status_code=403, detail="Missing `access_token`")
+
+    # if access_token == `token` then OK
+    if access_token != api_settings.global_access_token:
+        raise HTTPException(status_code=403, detail="Invalid `access_token`")
```
vincentsarago, line 68: should the exception be `401` (`UNAUTHORIZED`) here?

DeflateAwning: In the code examples at `\titiler\docs\src\examples\code\tiler_with_auth.md` and `\titiler\docs\src\intro.md`, they use 403. I agree that 401 is optimal though. I'll change it everywhere.

DeflateAwning:
Fixed. I changed 403 to 401 everywhere applicable (including docs).

src/titiler/application/titiler/application/settings.py
```diff
@@ -20,6 +22,9 @@ class ApiSettings(BaseSettings):
 
     lower_case_query_parameters: bool = False
 
+    # an API key required to access any endpoint, passed via the ?access_token= query parameter
+    global_access_token: Union[str, None] = None
```
vincentsarago, line 26:
```suggestion
    global_access_token: Optional[str] = None
```

src/titiler/application/titiler/application/settings.py
```diff
@@ -1,5 +1,7 @@
 """Titiler API settings."""
 
+from typing import Union
```
vincentsarago, line 3:
```suggestion
from typing import Optional
```

PR #749, opened 2024-01-10
Fix leafletjs template maxZoom to great than 18 zoom

Firefishy:
## What I am changing
Fix the leaflet map template to allow zooming beyond zoom 18.

## How I did it
The tilelayer incorrectly set `maxNativeZoom` instead of `maxZoom`. The `maxZoom` default is 18 and would not load tiles beyond 18 regardless of the layer's zoom setting.

## How you can test it
Tested in browser with correct setting.


vincentsarago:
thanks @Firefishy 🙏

PR #740, opened 2023-12-12
Add build support for arm64

Firefishy:
THIS IS FOR COMMENT. Needs some clean-up / refactor work.

## What I am changing
- ARM64 build of the gunicorn container
- Update GitHub Actions to latest releases
- Simplify the container tagging by using `docker/metadata-action` (also fix a case sensitive bug on fork)
- Add if statements to allow actions to cleanly run if forked
- Add dependabot to keep GHA updated

## How I did it
- TBC

## How you can test it
- https://github.com/Firefishy/titiler/actions/runs/7186402001/job/19571794840

## Related Issues
- https://github.com/developmentseed/titiler/discussions/387

PR #764, opened 2024-01-31
add dependabot

chris-bateman:
This is the long term fix for the github action updates.
Adjust to add limits etc to reduce dependabot spam.


chris-bateman:
Just saw #740 contains the same.  So happy with either PR.

PR #771, opened 2024-02-01
Update custom tms code example

GeoWill:
Hiya, 
Just playing around with custom tile matrix sets, and had to change the example from the docs to get it working. 
This is what I did, but still have quite a shallow understanding, so might not be the best way. Happy to alter if needed.


vincentsarago:
🙏 Thank you

PR #796, opened 2024-03-14
add colormap endpoints

vincentsarago:
ref https://github.com/developmentseed/titiler/discussions/509

Screenshot 2024-03-14 at 4 08 31 PM

Screenshot 2024-03-14 at 4 09 20 PM

Screenshot 2024-03-14 at 4 09 35 PM

Screenshot 2024-03-14 at 4 10 10 PM

Screenshot 2024-03-14 at 4 22 20 PM

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.17.0
+current_version = 0.17.1
 commit = True
 tag = True
 tag_name = {new_version}
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.17.2
+current_version = 0.17.3
 commit = True
 tag = True
 tag_name = {new_version}
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,12 +1,12 @@
 [bumpversion]
-current_version = 0.17.1
+current_version = 0.17.2
 commit = True
 tag = True
 tag_name = {new_version}
-parse = 
+parse =
 	(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)
 	((?P<pre>a|b|rc)(?P<prenum>\d+))?
-serialize = 
+serialize =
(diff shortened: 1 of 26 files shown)


vincentsarago:
## TODO 

- [x] ask for feedback
- [ ] add tests

JinIgarashi:
This new endpoint looks very useful. I would suggest to have a query param for `number of classes` at the endpoint of `/colormap/{id}`.
I can see current endpoint returns so many classes (44 No. what I can see in screenshot), but it would be nice if client side can request how many classes to be returned from titiler. I think `number of classes` parameter can be useful if want to use custom colormap to classify data with limited classes instead of linear legend.

vincentsarago:
@JinIgarashi thank for your feedback 🙏 

> (44 No. what I can see in screenshot)

in fact it returns 256 values, because all rio-tiler colormaps are composed of values between 0 and 255 (because that's how GDAL colormap are)

We could add a `step` parameter to only select few values but user would need to understand that for colormaps (e.g qualitative colormaps) this would give really weird results 

image

vincentsarago:
> Also, it might be useful if it can return color as hsl format in addition to rgba and hex formats. If color format is following [maplibre style spec](https://maplibre.org/maplibre-style-spec/types/#color), it can be easier to adapt for maplibre/mapbox.

Well the more I think about this, the more I think we should only return value as RGBA tuple, not event as HEX string. It seems quite trivial do transform the RGBA values to hex or hsl in client application instead of adding more complexity in the titiler endpoint

vincentsarago:
> This new endpoint looks very useful. I would suggest to have a query param for number of classes at the endpoint of /colormap/{id}.

same as https://github.com/developmentseed/titiler/pull/796#issuecomment-1999296498, this would be pretty trivial to do in the client application directly instead of adding complexity in titiler endpoint IMO

vincentsarago:
Q: we have the `tilematrixet` endpoints with a `/tileMatrixSet` prefix (defined by OGC), so should the colormap endpoints be prefixed with `colorMaps` (instead of `colormaps` as defined now)? 

```
## This PR 

/colormaps
/colormaps/{colormapId}

## Proposed

/colorMaps
/colorMaps/{colormapId}
```

Issue #799, opened 2024-03-19
Issue with pydantic-settings, Extra inputs are not permitted


imanshafiei540:
#### Problem description

I encountered an error related to the `rio_tiler_max_threads` environment variable when using TiTiler with the following environment configuration:

`.env` file:
```
RIO_TILER_MAX_THREADS=50
```

`pyproject.toml` file:
```toml
dependencies = [
    ...
    "pydantic-settings~=2.0",
]
```

`pip freeze` output:
```
pydantic-settings==2.2.1
```

The error occurred with the current implementation:
```python
model_config = SettingsConfigDict(env_prefix="TITILER_API_", env_file=".env")
```

The error message is:
```
rio_tiler_max_threads
  Extra inputs are not permitted [type=extra_forbidden, input_value='50', input_type=str]
    For further information visit https://errors.pydantic.dev/2.4/v/extra_forbidden
```

#### Expected Output

No errors when loading the environment variables.

#### Proposed Fix

I fixed the issue by allowing extra settings in the `SettingsConfigDict`:
```python
model_config = SettingsConfigDict(env_prefix="TITILER_API_", env_file=".env", extra="allow")
```

#### Environment Information

- Python Version: 3.11
- TiTiler Version: 0.17.1

vincentsarago:
@imanshafiei540 this is really weird because `RIO_TILER_MAX_THREADS` doesn't conflict with `TITILER_API_` prefix 🤷

imanshafiei540:
Okay, I tried to dig more into the issue and ended up here in this line from `Pydantic==2.6.4`:

```
self.__pydantic_validator__.validate_python(data, self_instance=self)
```

Somehow, the `api_settings = ApiSettings()` coming from `src/titiler/application/titiler/application/main.py` cannot be validated by `__pydantic_validator__` and I'm not sure what's the reason.

Could be some issues between `Pydantic` and `Pydantic-Settings.`

I hope someone else can also verify the issue; I'm constantly thinking maybe I'm doing something wrong on my end.

Anyway, thank you for the quick response, and I'll try to figure out what's wrong; it's probably time to raise an issue on the Pydantic-Setiings repo.

vincentsarago:
can you share the full `log`? and/or share a reproducible example 🙏

imanshafiei540:
Sure, I'll prepare something ASAP, just as a note, everything is working with `pydantic-settings==2.1.0`
`2.2.0` or above will break the TiTiler.

[...]

PR #800, opened 2024-03-20
Adding extra="ignore" option to fix pydantic_settings.SettingsConfigDict issue with environement variables

imanshafiei540:
Closes https://github.com/developmentseed/titiler/issues/799

Context:

`pydantic-settings` Released a new update (`2.2.0`) that breaks TiTiler API when using a `.env` file that contains variables inside.

This PR will solve the issue by adding `extra="ignore"` to the `pydantic_settings.SettingsConfigDict` object.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Release Notes
 
+## 0.17.3 (2024-03-21)
+
+### titiler.application
+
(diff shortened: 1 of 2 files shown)


vincentsarago:
🙏 thanks @imanshafiei540

Issue #801, opened 2024-03-21
Remove default TileMatrixSetId in `0.19` 


vincentsarago:
in order to align with the OGC Tiles specification, we have to remove the `default` tilematrixset in the endpoints 

```
# Now
/cog/tiles/{z}/{x}/{y} 

# Future
/cog/tiles/WebMercatorQuad/{z}/{x}/{y} 
```

This is quite a breaking change but it will makes things a bit more clean in the code IMO

ref #546

PR #802, opened 2024-03-21
deprecate default tilematrixset

vincentsarago:
ref #801 

This PR removes non-prefixed `/tileMatrixSetId` endpoints for `/tilejson.json`, `/tiles`, `/map` and `/WMTSCapabilities.xml`, which now makes `tileMatrixSetId` a required **path** parameter

```
# Before
/cog/tiles/{z}/{x}/{y} 
/cog/tilejson.json
/cog/map
/cog/WMTSCapabilities.xml

# Now
/cog/tiles/WebMercatorQuad/{z}/{x}/{y} 
/cog/WebMercatorQuad/tilejson.json
/cog/WebMercatorQuad/map
/cog/WebMercatorQuad/WMTSCapabilities.xml
```


src/titiler/core/titiler/core/factory.py
```diff
@@ -520,10 +520,18 @@ def geojson_statistics(
...
+            r"/tiles/{z}/{x}/{y}.{format}", **img_endpoint_params, deprecated=True
+        )
+        @self.router.get(
+            r"/tiles/{z}/{x}/{y}@{scale}x", **img_endpoint_params, deprecated=True
+        )
+        @self.router.get(
+            r"/tiles/{z}/{x}/{y}@{scale}x.{format}",
+            **img_endpoint_params,
+            deprecated=True,
+        )
```
vincentsarago, line 534: For now we just `deprecate` the endpoints but we will removes them in `0.19`

PR #815, opened 2024-05-06
fastapi-slim replaces fastapi

n8sty:
## What I am changing

Replace the `fastapi` dependency with `fastapi-slim`. 

See https://github.com/tiangolo/fastapi/pull/11522 for details. My brief summary of what happened in the FastAPI project: Installing `fastapi` now comes along with dependencies that are helpful for development and deployment, at the expense of installing more dependencies. Since tiler applications are opinionated in their dependencies additional dependencies raise the risk of conflicts, albeit slightly. This change keeps the same dependencies that FastAPI pulls in transitively.

## How I did it

Swap the dependency specification, meaning nothing would change :smile:.

## How you can test it

Unit tests, standard CI should be more than enough because this result in nothing changing.

## Related Issues

N/A.

### DIFF
--- a/src/titiler/core/pyproject.toml
+++ b/src/titiler/core/pyproject.toml
@@ -31,7 +31,7 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
-    "fastapi>=0.107.0",
+    "fastapi-slim",
     "geojson-pydantic>=1.0,<2.0",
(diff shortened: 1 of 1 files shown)


vincentsarago:
such a shame, I don't really get why they decided to change this 😬 

@n8sty thanks for the PR 🙏

vincentsarago (approved):

PR #819, opened 2024-05-25
remove defaults in deps and add `.as_dict` method to the `DefaultDependency` class

vincentsarago:
This PR does:
- remove defaults to most of dependencies (and default to rio-tiler values)
- In order to defaults to rio-tiler's default we have to NOT pass options to rio-tiler's method. when we use Dependency unpacking we would pass `value=None`, to overcome this, we introduce a new `.as_dict()` method which will return a Dict from the dependency but only with non-None values


vincentsarago:
this PR is a draft because I need to double check all the implication of the `.kwargs` property

PR #987, opened 2024-09-20
feature: add links for render parameters and /cog/map link to /cog/viewer dashboard

hrodmn:
If someone is using the `/cog/viewer` endpoint to figure out which render parameters make sense for their dataset it would be nice to add a `Show Render Parameters` button. It would also be nice to be able to share a link to a map with those parameters applied so the user could send someone a link to the exact image that they decided on. It might be better to link to `/cog/viewer` since `/cog/map` is deprecated, but right now you can't set render parameters in that endpoint (just `url`).

image

This idea came from Team 4 in the IMPACT X-athon with @anayeaye    @smk0033 @unhipcat

### DIFF
--- a/src/titiler/extensions/titiler/extensions/templates/cog_viewer.html
+++ b/src/titiler/extensions/titiler/extensions/templates/cog_viewer.html
@@ -955,8 +955,7 @@ document.getElementById('btn-share-link').addEventListener('click', () => {
     params.append('rescale', `${document.getElementById("data-min").value},${document.getElementById("data-max").value}`);
   }
 
-  const path_name = `${window.location.pathname}`.replace("viewer", "map");
-  console.log("path_name: ", path_name);
+  const path_name = `${window.location.pathname}`.replace("viewer", "WebMercatorQuad/map");
   const shareUrl = `${window.location.origin}${path_name}?${params.toString()}`;
   
   // Create a temporary input element to copy the URL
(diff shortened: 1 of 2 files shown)


vincentsarago:
🥳 
thanks for starting this @hrodmn 

Would you be able to change few things, mostly making `list` parameters as an array `[]` in the render object:
- `"rescale": "0,1000"` -> `"rescale": [[0,1000]]`
- `"bidx": 1` -> `"bidx": [1]`

vincentsarago:
also `/cog/map` is deprecated but not `/cog/{tileMatrixSetId}/map` ;-)

hrodmn:
> We could update the code to accept more parameters ... but to be honest I would prefer if we re-write the whole application 😅

It would be great to move away from the giant `html` template file someday, but it works!

src/titiler/extensions/titiler/extensions/templates/cog_viewer.html
```diff
@@ -955,8 +955,7 @@
     params.append('rescale', `${document.getElementById("data-min").value},${document.getElementById("data-max").value}`);
   }
 
-  const path_name = `${window.location.pathname}`.replace("viewer", "map");
-  console.log("path_name: ", path_name);
+  const path_name = `${window.location.pathname}`.replace("viewer", "WebMercatorQuad/map");
```
vincentsarago, line 958: 😓 the only issue is that the `/map` endpoint might not be available

vincentsarago: yes

hrodmn: I'll figure out how to make it conditionally visible.

vincentsarago: @hrodmn because the viewer is an extension, when we create the endpoint we have access to the factory attributes so in https://github.com/developmentseed/titiler/blob/4327ab83daeed7a17ba91d0f761d13c76c3a301d/src/titiler/extensions/titiler/extensions/viewer.py#L24-L42 you could pass a boolean in the template context

src/titiler/extensions/titiler/extensions/viewer.py
```diff
@@ -36,6 +36,8 @@ def cog_viewer(request: Request):
                     ),
                     "info_endpoint": factory.url_for(request, "info"),
                     "statistics_endpoint": factory.url_for(request, "statistics"),
+                    "viewer_enabled": hasattr(factory, "add_viewer")
```
vincentsarago, line 39: so if the attribute is there but set to False what happens?

hrodmn: If it is set to `False` the `Share Map` button is hidden in the webmap:
image

The check for `getattr(factory, "add_viewer", False)` is on the next line but is not obvious with `black` formatting :smile:

src/titiler/extensions/titiler/extensions/viewer.py
```diff
@@ -36,6 +36,8 @@ def cog_viewer(request: Request):
...
+                    and getattr(factory, "add_viewer", False),
```
vincentsarago, line 40: I think that `getattr(class, "name", default)` should be enough 

> If the named attribute does not exist, default is returned if provided, otherwise [AttributeError](https://docs.python.org/3/library/exceptions.html#AttributeError) is raised. name need not be a Python identifier (see [setattr()](https://docs.python.org/3/library/functions.html#setattr)).

PR #995, opened 2024-10-03
Add missing supported_crs to mosaicjson wmts endpoint

AndrewAnnex:
also made the bounds_crs discovered from the tms's geographic crs. This enables non-earth TMSs to work with GDAL/QGIS

currently I call '.srs' on the geographic crs to get the bounds crs but it may make sense to add a function to determine the more compact urn representation first, failing that allow the WKT representation to pass through.

I was able to sort of test this by manually editing the WMTS response until it worked with QGIS but let me think about the srs question above a bit more before merging. As far as I could tell, QGIS and GDAL had no issue parsing a WKT2 crs string for the supported_crs but I am not exactly sure if that is true also for the bounding box crs. 

It will be difficult to test with non earth TMSs within this repository as the ones I am developing are not official with the naming authority so this may take a few tries. At least it shouldn't break any existing functionality or tests..

### DIFF
--- a/src/titiler/core/titiler/core/factory.py
+++ b/src/titiler/core/titiler/core/factory.py
@@ -862,6 +862,8 @@ class TilerFactory(BaseFactory):
                 supported_crs = f"EPSG:{tms.crs.to_epsg()}"
             else:
                 supported_crs = tms.crs.srs
+            
+            bounds_crs = tms.crs.geographic_crs.srs or "urn:ogc:def:crs:OGC:2:84"
 
             return self.templates.TemplateResponse(
                 request,
@@ -869,6 +871,7 @@ class TilerFactory(BaseFactory):
                 context={
                     "tiles_endpoint": tiles_url,
                     "bounds": bounds,
+                    "bounds_crs": bounds_crs,
                     "tileMatrix": tileMatrix,
                     "tms": tms,
                     "supported_crs": supported_crs,
--- a/src/titiler/core/titiler/core/factory.py
+++ b/src/titiler/core/titiler/core/factory.py
@@ -574,7 +574,10 @@ class TilerFactory(BaseFactory):
             tms = self.supported_tms.get(tileMatrixSetId)
             with rasterio.Env(**env):
                 with self.reader(
-                    src_path, tms=tms, **reader_params.as_dict()
+                    src_path,
+                    tms=tms,
+                    geographic_crs=tms.geographic_crs,
+                    **reader_params.as_dict(),
                 ) as src_dst:
                     image = src_dst.tile(
                         x,
@@ -684,7 +687,10 @@ class TilerFactory(BaseFactory):
             tms = self.supported_tms.get(tileMatrixSetId)
             with rasterio.Env(**env):
                 with self.reader(
-                    src_path, tms=tms, **reader_params.as_dict()
+                    src_path,
+                    tms=tms,
+                    geographic_crs=tms.geographic_crs,
+                    **reader_params.as_dict(),
                 ) as src_dst:
                     return {
                         "bounds": src_dst.geographic_bounds,
@@ -838,7 +844,10 @@ class TilerFactory(BaseFactory):
             tms = self.supported_tms.get(tileMatrixSetId)
             with rasterio.Env(**env):
                 with self.reader(
-                    src_path, tms=tms, **reader_params.as_dict()
+                    src_path,
+                    tms=tms,
+                    geographic_crs=tms.geographic_crs,
+                    **reader_params.as_dict(),
                 ) as src_dst:
                     bounds = src_dst.geographic_bounds
                     minzoom = minzoom if minzoom is not None else src_dst.minzoom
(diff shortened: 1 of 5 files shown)


src/titiler/core/titiler/core/factory.py
```diff
@@ -862,13 +862,16 @@ def wmts(
                 supported_crs = f"EPSG:{tms.crs.to_epsg()}"
             else:
                 supported_crs = tms.crs.srs
+            
+            bounds_crs = tms.crs.geographic_crs.srs or "urn:ogc:def:crs:OGC:2:84"
```
vincentsarago, line 866:
```suggestion
            bounds_crs = tms.crs.geographic_crs.srs or "urn:ogc:def:crs:OGC:2:84"
```

did you meant `geodetic_crs` instead of `geographic_crs` ? or maybe `tms.geographic_crs` ?

the issue is that the `srs` is not a `uri` 

```
In [4]: pyproj.CRS.from_epsg(4326).geodetic_crs.srs
Out[4]: 'GEOGCRS["WGS 84",ENSEMBLE["World Geodetic System 1984 ensemble",MEMBER["World Geodetic System 1984 (Transit)"],MEMBER["World Geodetic System 1984 (G730)"],MEMBER["World Geodetic System 1984 (G873)"],MEMBER["World Geodetic System 1984 (G1150)"],MEMBER["World Geodetic System 1984 (G1674)"],MEMBER["World Geodetic System 1984 (G1762)"],MEMBER["World Geodetic System 1984 (G2139)"],ELLIPSOID["WGS 84",6378137,298.257223563,LENGTHUNIT["metre",1]],ENSEMBLEACCURACY[2.0]],PRIMEM["Greenwich",0,ANGLEUNIT["degree",0.0174532925199433]],CS[ellipsoidal,2],AXIS["geodetic latitude (Lat)",north,ORDER[1],ANGLEUNIT["degree",0.0174532925199433]],AXIS["geodetic longitude (Lon)",east,ORDER[2],ANGLEUNIT["degree",0.0174532925199433]],USAGE[SCOPE["Horizontal component of 3D system."],AREA["World."],BBOX[-90,-180,90,180]],ID["EPSG",4326]]'
```

src/titiler/core/titiler/core/templates/wmts.xml
```diff
@@ -37,7 +37,7 @@
             <ows:Title>{{ title }}</ows:Title>
             <ows:Identifier>{{ layer_name }}</ows:Identifier>
             <ows:Abstract>{{ title }}</ows:Abstract>
-            <ows:WGS84BoundingBox crs="urn:ogc:def:crs:OGC:2:84">
+            <ows:WGS84BoundingBox crs={{ bounds_crs }}>
```
vincentsarago, line 40: would be weird to have `WGS84BoundingBox` not being in WGS84

src/titiler/core/titiler/core/factory.py
AndrewAnnex: ahh yes `tms.geographic_crs` is the right call here. I think I also confirmed locally that the bound's crs must be a urn so will look to change that.

src/titiler/core/titiler/core/templates/wmts.xml
AndrewAnnex: it is but I think this is just an artifact of the the WMTS spec, GDAL (via gdalinfo/qgis) seemed happy to work with a mars urn here in the crs. Unless its possible to call it a `BoundingBox` sans geoid name? I'll have to dig more into the wmts spec to see if there is a different choice to make.

AndrewAnnex: ahh looks like OGC 06-121r3 does mention an alternative `BoundingBox` that is more extendable (can be n dimensional for example). Now I can either update to only use `BoundingBox` or use `WGS84BoundingBox` if the bounds_crs is 4326. thoughts?

AndrewAnnex:
@vincentsarago, I noticed that when trying to make a WMTS from a single cog that the geographic_crs bounds were getting set to the whole globe, so grabbing the geographic_crs from the TMS I think makes sense to do here without needing to update rio_tiler, so I updated the pr to be more explicit about that.

I think however that the coord_crs may need to be added as a query parameter to `/info` and `/info.geojson` and similar functions where the no TMS is assumed, but the correct geographic_crs should be discoverable from the cog itself.

vincentsarago:
> I think however that the coord_crs may need to be added as a query parameter to /info and /info.geojson and similar functions where the no TMS is assumed, but the correct geographic_crs should be discoverable from the cog itself.

I feel I should remove the geographic bounds and info about min/max zoom in `/info` and let that info only in the /tilejson.json endpoint

vincentsarago:
@AndrewAnnex the issue is that having tms related information in `/info` makes things way more complicated that it should be. 

same for the `bounds`, I'm not convinced that having `bounds` in the `/info` response is used and also makes sense without returning the CRS.

I'm going to push a PR in rio-tiler explaining the changes.

vincentsarago:
done in #1004

PR #1003, opened 2024-10-22
update rio-tiler dependency

vincentsarago:
This PR does:

- update dependencies: rio-tiler, cogeo-mosaic, geojson-pydantic
- better handle antimeridian crossing dataset in `/info.geojson` 
- use `/info.geojson` endpoints in viewers

- add `crs` options in `/bounds` and `/info.geojson` endpoints for users to set their own geographic-crs


src/titiler/core/titiler/core/factory.py
```diff
@@ -320,27 +320,6 @@ def register_routes(self):
...
-        )
-        def bounds(
-            src_path=Depends(self.path_dependency),
-            reader_params=Depends(self.reader_dependency),
-            env=Depends(self.environment_dependency),
-        ):
-            """Return the bounds of the COG."""
-            with rasterio.Env(**env):
-                with self.reader(src_path, **reader_params.as_dict()) as src_dst:
-                    return {"bounds": src_dst.geographic_bounds}
```
vincentsarago, line 342: The `bounds` endpoints isn't really useful alone. Users can use the `/tilejson.json` or `/info.geojson` to get more geo information about the dataset 

Note: `geographic_bounds` has been remove in rio-tiler and replaced with `get_geographic_bounds`

vincentsarago: changed my mind 🙈

src/titiler/mosaic/titiler/mosaic/factory.py
```diff
@@ -237,11 +245,22 @@ def info_geojson(
                     reader_options=reader_params.as_dict(),
                     **backend_params.as_dict(),
                 ) as src_dst:
-                    info = src_dst.info()
+                    bounds = src_dst.get_geographic_bounds(crs or WGS84_CRS)
+                    if bounds[0] > bounds[2]:
+                        pl = Polygon.from_bounds(-180, bounds[1], bounds[2], bounds[3])
```
AndrewAnnex, line 250: wondering how safe this is to assume, if dealing with a pyproj CRS and the area of use is defined that could be one way to get this, but I wonder for other codes and coordinate system axis orders... Probably something though that isn't super important and can be solved later

AndrewAnnex: oh wait you get the bounds first, okay then this is porbably fine as xy order is enforced

vincentsarago: 👍 good point, let's wait until someone gets the issue 😅 

For now it will cover 98% of the users I think

PR #1004, opened 2024-10-22
use BoundingBox instead of WGS84BoundingBox in wmts when tms is not based on WGS84

vincentsarago:
second take at #995 

cc @AndrewAnnex

### DIFF
--- a/src/titiler/mosaic/titiler/mosaic/factory.py
+++ b/src/titiler/mosaic/titiler/mosaic/factory.py
@@ -590,6 +590,12 @@ class MosaicTilerFactory(BaseFactory):
                 Optional[int],
                 Query(description="Overwrite default maxzoom."),
             ] = None,
+            use_epsg: Annotated[
+                bool,
+                Query(
+                    description="Use EPSG code, not opengis.net, for the ows:SupportedCRS in the TileMatrixSet (set to True to enable ArcMap compatability)"
+                ),
+            ] = False,
             layer_params=Depends(self.layer_dependency),
             dataset_params=Depends(self.dataset_dependency),
             pixel_selection=Depends(self.pixel_selection_dependency),
@@ -621,6 +627,7 @@ class MosaicTilerFactory(BaseFactory):
                 "minzoom",
                 "maxzoom",
                 "service",
+                "use_epsg",
                 "request",
             ]
             qs = [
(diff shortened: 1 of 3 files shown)


src/titiler/mosaic/titiler/mosaic/factory.py
```diff
@@ -590,6 +590,12 @@ def wmts(
                 Optional[int],
                 Query(description="Overwrite default maxzoom."),
             ] = None,
+            use_epsg: Annotated[
+                bool,
+                Query(
+                    description="Use EPSG code, not opengis.net, for the ows:SupportedCRS in the TileMatrixSet (set to True to enable ArcMap compatability)"
+                ),
+            ] = False,
```
vincentsarago, line 598: not related to the original PR

src/titiler/mosaic/titiler/mosaic/factory.py
```diff
@@ -659,6 +666,17 @@ def wmts(
...
+            if use_epsg:
+                supported_crs = f"EPSG:{tms.crs.to_epsg()}"
+            else:
+                supported_crs = tms.crs.srs
+
+            bbox_crs_type = "WGS84BoundingBox"
+            bbox_crs_uri = "urn:ogc:def:crs:OGC:2:84"
+            if tms.rasterio_geographic_crs != WGS84_CRS:
+                bbox_crs_type = "BoundingBox"
+                bbox_crs_uri = CRS_to_uri(tms.rasterio_geographic_crs)
```
AndrewAnnex, line 678: here an above, I think that uri's are not the same as urn's (functionally very similar) although I forget if qgis/gdal was fine with either representation. I also made this code addition as a feature branch a while ago to morecantile that could be used but don't want to block this pr https://github.com/developmentseed/morecantile/pull/163/files

vincentsarago: to be honest I've read the WMTS spec and it's not clear what the `URI` should be, I feel it would be fine as long that pyproj/rasterio can resolve `CRS.from_user_input(uri)`

AndrewAnnex: okay great https://github.com/cogeotiff/rio-tiler/pull/752 is up, will need to look again at https://github.com/rasterio/rasterio/pull/3206 soon though to get my datasets supported though (hopefully would be a patch level fix...)

PR #1007, opened 2024-10-24
[WIP] add `titiler.xarray` module

vincentsarago:
ref: https://github.com/developmentseed/titiler-xarray/issues/68

depends on https://github.com/cogeotiff/rio-tiler/pull/755

## To Do 
- [ ] review `.io.Reader`
- [ ] add Notebook for the new Reader
- [ ] write factory
- [ ] add tests
- [ ] add docs
- [ ] update CI to release titiler.xarray
- [ ] add example of how to use the xarray.io.Reader in STAC based tiler


src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,273 @@
...
+
+@attr.s
+class Reader(XarrayReader):
+    """Reader: Open Zarr file and access DataArray."""
+
+    src_path: str = attr.ib()
+    variable: str = attr.ib()
+
+    # xarray.Dataset options
+    opener: Callable[..., xarray.Dataset] = attr.ib(default=xarray_open_dataset)
```
vincentsarago, line 212: By making the `opener` an option it eases the user to bring their own `xarray_open_dataset` callable which may add more `protocol` (e.g Azure, GCS ...)

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,273 @@
...
+    """Get Xarray variable as DataArray."""
+    da = ds[variable]
+    da = arrange_coordinates(da)  # TODO: Duplicated call L173
+
+    # TODO: add test
+    if drop_dim:
+        dim_to_drop, dim_val = drop_dim.split("=")
+        da = da.sel({dim_to_drop: dim_val}).drop(dim_to_drop)
+
+    da = arrange_coordinates(da)
```
vincentsarago, line 177: from titiler-xarray repo: we have twice a call to arrange_coordinates... not sure why cc @abarciauskas-bgse

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,273 @@
...
+    variable: str = attr.ib()
+
+    # xarray.Dataset options
+    opener: Callable[..., xarray.Dataset] = attr.ib(default=xarray_open_dataset)
+
+    group: Optional[Any] = attr.ib(default=None)
+    reference: bool = attr.ib(default=False)
+    decode_times: bool = attr.ib(default=False)
+    consolidated: Optional[bool] = attr.ib(default=True)
+    cache_client: Optional[CacheClient] = attr.ib(default=None)
```
vincentsarago, line 218: Cache client is not optional, basically users just need to provide an Object with `get` and `set` methods

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,280 @@
...
+
+    """
+    if "x" not in da.dims and "y" not in da.dims:
+        latitude_var_name = next(
+            x for x in ["lat", "latitude", "LAT", "LATITUDE"] if x in da.dims
+        )
+        longitude_var_name = next(
+            x for x in ["lon", "longitude", "LON", "LONGITUDE"] if x in da.dims
+        )
+        da = da.rename({latitude_var_name: "y", longitude_var_name: "x"})
```
vincentsarago, line 156: TODO: add a test if `latitude_var_name` or `longitude_var_name` are None

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,280 @@
...
+
+    if "TIME" in da.dims:
+        da = da.rename({"TIME": "time"})
```
vincentsarago, line 159: this was added for 

```python
from titiler.xarray.io import Reader

with Reader("s3://aodn-cloud-optimised/model_sea_level_anomaly_gridded_realtime.zarr/", "GSL") as src:
   print(src.info())

bounds=(-180.28109374046326, -60.1, 180.081081533432, 10.1) crs='http://www.opengis.net/def/crs/EPSG/0/4326' band_metadata=[] band_descriptions=[] dtype='float64' nodata_type='Nodata' colorinterp=None scales=None offsets=None colormap=None name='GSL' count=1 width=641 height=351 attrs={'description': 'GSLA + CAST2008 mean dynamic height ', 'long_name': 'gridded sea level', 'standard_name': 'sea_surface_height_above_geoid', 'units': 'm'}

```

src/titiler/xarray/titiler/xarray/io.py
abarciauskas-bgse: 🤔 seems like an error, we can probably safely remove it in line 170

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,284 @@
...
+    if "time" in da.dims:
+        da = da.transpose("time", "y", "x")
+    else:
+        da = da.transpose("y", "x")
+
+    # If min/max values are stored in `valid_range` we add them in `valid_min/valid_max`
+    vmin, vmax = da.attrs.get("valid_min"), da.attrs.get("valid_max")
+    if "valid_range" in da.attrs and not (vmin is not None and vmax is not None):
+        valid_range = da.attrs.get("valid_range")
+        da.attrs.update({"valid_min": valid_range[0], "valid_max": valid_range[1]})
```
vincentsarago, line 170: not sure if `valid_range` is something we find usually

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,289 @@
...
+        )
+        longitude_var_name = next(
+            x for x in ["lon", "longitude", "LON", "LONGITUDE", "Lon"] if x in da.dims
+        )
+        da = da.rename({latitude_var_name: "y", longitude_var_name: "x"})
+
+    if "TIME" in da.dims:
+        da = da.rename({"TIME": "time"})
+
+    if _dims := [d for d in da.dims if d not in ["x", "y"]]:
```
vincentsarago, line 161: This is to support dataset with more than 3 dimension (e.g more than time, x, y)

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,289 @@
...
+    da = da.rio.write_crs(crs)
+
+    if crs == "epsg:4326" and (da.x > 180).any():
+        # Adjust the longitude coordinates to the -180 to 180 range
+        da = da.assign_coords(x=(da.x + 180) % 360 - 180)
+
+        # Sort the dataset by the updated longitude coordinates
+        da = da.sortby(da.x)
+
+    # TODO: Technically we don't have to select the first time, rio-tiler should handle 3D dataset
```
vincentsarago, line 201: ☝️ we might enable all the `times` values, and not select the first value by default

vincentsarago:
Screenshot 2024-10-29 at 6 47 50 PM

src/titiler/core/titiler/core/factory.py
```diff
@@ -885,6 +883,16 @@ def wmts(
...
 
+            layers = [
+                {
+                    "title": src_path if isinstance(src_path, str) else "TiTiler",
+                    "name": "default",
+                    "tiles_url": tiles_url,
+                    "query_string": urlencode(qs, doseq=True) if qs else None,
+                    "bounds": bounds,
+                },
+            ]
```
vincentsarago, line 894: add layers to WMTS response 

this will be use for STAC renders but maybe also for multiple variable support in XARRAY

src/titiler/core/titiler/core/templates/map.html
```diff
@@ -86,20 +86,33 @@
...
+      bottom = bounds[1],
+      right = bounds[2],
+      top = bounds[3];
+
+    if (left < right) {
+      left = Math.max(left, {{ tms.bbox.left }})
+      right = Math.min(right, {{ tms.bbox.right }})
+    }
+    bottom = Math.max(bottom, {{ tms.bbox.bottom }})
+    top = Math.min(top, {{ tms.bbox.top }})
```
vincentsarago, line 99: I needed to do this because some NetCDF data have bounds that exceed the TMS bbox

src/titiler/xarray/titiler/xarray/dependencies.py
```diff
@@ -0,0 +1,211 @@
...
+
+@dataclass
+class XarrayParams(XarrayIOParams, XarrayDsParams):
+    """Xarray Reader dependency."""
+
+    pass
+
+
+@dataclass(init=False)
+class CompatXarrayParams(DefaultDependency):
```
vincentsarago, line 77: this is not directly used in titiler.xarray but could be in a Tiler that would want to support both GDAL/Xarray dataset

src/titiler/xarray/titiler/xarray/factory.py
```diff
@@ -0,0 +1,364 @@
...
+    PartFeatureParams,
+    TileParams,
+    XarrayIOParams,
+    XarrayParams,
+)
+from titiler.xarray.io import Reader
+
+
+@define(kw_only=True)
+class TilerFactory(BaseTilerFactory):
```
vincentsarago, line 42: By sub-classing `titiler.core.factory.TilerFactory` we avoid re-writing code

src/titiler/xarray/titiler/xarray/factory.py
```diff
@@ -0,0 +1,364 @@
...
+
+    # Custom dependency for /variable
+    io_dependency: Type[DefaultDependency] = XarrayIOParams
+
+    add_viewer: bool = True
+    add_part: bool = True
+
+    # remove some attribute from init
+    img_preview_dependency: Type[DefaultDependency] = field(init=False)
+    add_preview: bool = field(init=False)
```
vincentsarago, line 74: we don't want to have `preview` for now (I don't think any user would want this except if they control the dataset available)

src/titiler/xarray/titiler/xarray/factory.py
```diff
@@ -0,0 +1,364 @@
...
+            content, media_type = render_image(
+                image,
+                output_format=format,
+                colormap=colormap,
+                **render_params.as_dict(),
+            )
+
+            return Response(content, media_type=media_type)
+
+    # custom /statistics endpoints (remove /statistics - GET)
```
vincentsarago, line 293: ☝️ IMO having a full dataset /statistics in a bit dangerous (as for the /preview endpoints)

src/titiler/xarray/titiler/xarray/factory.py
```diff
@@ -0,0 +1,364 @@
...
+                    "content": {"application/geo+json": {}},
+                    "description": "Return dataset's statistics from feature or featureCollection.",
+                }
+            },
+        )
+        def geojson_statistics(
+            geojson: Annotated[
+                Union[FeatureCollection, Feature],
+                Body(description="GeoJSON Feature or FeatureCollection."),
+            ],
```
vincentsarago, line 314: NOTE: we should maybe extract this and make a dependency that user could customize to control the size of the GeoJSON

src/titiler/xarray/titiler/xarray/main.py
```diff
@@ -0,0 +1,30 @@
...
+    """,
+)
+
+md = TilerFactory(router_prefix="/md")
+
+app.include_router(
+    md.router,
+    prefix="/md",
+    tags=["Multi Dimensional"],
+)
```
vincentsarago, line 30: we might remove this in the final module

PR #1013, opened 2024-10-29
Add Xarray sub-package 

vincentsarago:
overtake #1007 

ref: https://github.com/developmentseed/titiler-xarray/issues/68

### To Do

- [x] reviews
- [x] add Docs


src/titiler/xarray/titiler/xarray/main.py
```diff
@@ -0,0 +1,30 @@
...
+    """,
+)
+
+md = TilerFactory(router_prefix="/md")
+
+app.include_router(
+    md.router,
+    prefix="/md",
+    tags=["Multi Dimensional"],
+)
```
vincentsarago, line 30: TODO: remove this before merging 

`titiler.xarray` should be seen as a plug-in to titiler and not as an application itself. We will add example on how to build application using the endpoint factory

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,306 @@
...
+    An rioxarray.exceptions.InvalidDimensionOrder error is raised if the coordinates are not in the correct order time, y, and x.
+    See: https://github.com/corteva/rioxarray/discussions/674
+
+    We conform to using x and y as the spatial dimension names..
+
+    """
+    if "x" not in da.dims and "y" not in da.dims:
+        try:
+            latitude_var_name = next(
+                x for x in ["lat", "latitude", "LAT", "LATITUDE", "Lat"] if x in da.dims
```
vincentsarago, line 152: do we need to support other variable name?

src/titiler/xarray/titiler/xarray/dependencies.py
```diff
@@ -0,0 +1,211 @@
...
+
+@dataclass
+class XarrayParams(XarrayIOParams, XarrayDsParams):
+    """Xarray Reader dependency."""
+
+    pass
+
+
+@dataclass(init=False)
+class CompatXarrayParams(DefaultDependency):
```
vincentsarago, line 77: this is not directly used in titiler.xarray but could be in a Tiler that would want to support both GDAL/Xarray dataset

src/titiler/xarray/titiler/xarray/factory.py
```diff
@@ -0,0 +1,364 @@
...
+    PartFeatureParams,
+    TileParams,
+    XarrayIOParams,
+    XarrayParams,
+)
+from titiler.xarray.io import Reader
+
+
+@define(kw_only=True)
+class TilerFactory(BaseTilerFactory):
```
vincentsarago, line 42: By sub-classing titiler.core.factory.TilerFactory we avoid re-writing code

src/titiler/xarray/titiler/xarray/factory.py
```diff
@@ -0,0 +1,364 @@
...
+
+    # Custom dependency for /variable
+    io_dependency: Type[DefaultDependency] = XarrayIOParams
+
+    add_viewer: bool = True
+    add_part: bool = True
+
+    # remove some attribute from init
+    img_preview_dependency: Type[DefaultDependency] = field(init=False)
+    add_preview: bool = field(init=False)
```
vincentsarago, line 74: we remove those 2 attributes because we don't support `/preview` endpoints

src/titiler/xarray/pyproject.toml
```diff
@@ -0,0 +1,70 @@
...
+    "h5netcdf",
+    "xarray",
+    "rioxarray",
+    "zarr",
+    "fsspec",
+    "s3fs",
+    "aiohttp",
+    "pandas",
+    "httpx",
+]
```
vincentsarago, line 43: todo: update rio-tiler to >=7.1

src/titiler/xarray/titiler/xarray/factory.py
```diff
@@ -0,0 +1,364 @@
...
+                image,
+                output_format=format,
+                colormap=colormap,
+                **render_params.as_dict(),
+            )
+
+            return Response(content, media_type=media_type)
+
+    # custom /statistics endpoints (remove /statistics - GET)
+    def statistics(self):
```
vincentsarago, line 294: ☝️ IMO having a full dataset /statistics in a bit dangerous (as for the /preview endpoints)

vincentsarago:
you can try this with

```
uvicorn titiler.xarray.main:app --port 8080 --reload
```

Screenshot 2024-10-29 at 10 28 02 PM

vincentsarago:
⚠️ there is an issue with Zarr pyramid (used with `multiscale=True` option)

When reading the file without `group=` option, we don't have access to any variable:

```python
from titiler.xarray.io import Reader

with Reader("tests/fixtures/pyramid.zarr", variable="dataset") as src:
    print(src)
...
KeyError: 'dataset'
...
KeyError: "No variable named 'dataset'. Variables on the dataset include []"
```
we need to pass a `group` BUT for `multiscale` usage we don't want to pass the group 🤷 it should be defined by the zoom level dynamically

I feel `multiscale` support is a bit experimental at this point and maybe we can remove it in the first iteration 

cc @abarciauskas-bgse @maxrjones

src/titiler/xarray/tests/fixtures/pyramid.zarr/.zgroup
```diff
@@ -0,0 +1,3 @@
+{
+    "zarr_format": 2
```
maxrjones, line 2: would be good to do a `git rm -r` on `src/titiler/xarray/tests/fixtures/pyramid.zarr` since pyramid support has been removed

maxrjones (commented):
Thanks for all your work here, @vincentsarago!

This is a very opinionated take, but I think titiler-xarray would be best off with two separate routes, each with its own set of optional dependencies. The first route would be `zarr`, which would open Zarr and virtual Zarr datasets using `xarray.open_zarr`. The second route would be `md`, which would opening any dataset readable by `xarray.open_dataset`.

The primary reason I think we should do this is that it would enable us to incentivize virtualizing datasets into zarr, which would lead to much faster tile generation. We could do this by:
1. Having all query parameters in the zarr route only relevant for `open_zarr`, simplifying API usage.
2. Automatically detect virtual datasets, removing the need for the `reference` parameter.
3. Lightening the image size for titiler-xarray deployments only using zarr because other readers would not be installed (and eventually obstore and/or icechunk could be used instead of the fsspec dependencies)

This would also simplify non-zarr usage for the following reasons:
1. Zarr specific parameters (e.g., group, consolidated) would not be included in endpoints in the `md` route
2. We could use xarray's automatic backend detection rather than writing our own in  `titiler/xarray/io.py`

I also think isolating Zarr usage would simplify the eventual support of the GeoZarr and multiscales specifications.

src/titiler/xarray/tests/fixtures/pyramid.zarr/.zgroup
vincentsarago: I still test the `group` option for now

vincentsarago:
Thanks @maxrjones 🙏 

I see what you're saying. The goal of having a single Reader was to handle all the non-COG dataset so splitting in to two separate reader/set of endpoints would not meat the goal. 

> This is a very opinionated take, but I think titiler-xarray would be best off with two separate routes, each with its own set of optional dependencies. The first route would be zarr, which would open Zarr and virtual Zarr datasets using xarray.open_zarr. The second route would be md, which would opening any dataset readable by xarray.open_dataset.

We can absolutely use `xarray.open_zarr` instead of `xarray.open_dataset` here when reading a zarr

> We could use xarray's automatic backend detection rather than writing our own in  titiler/xarray/io.py

How so? https://github.com/developmentseed/titiler/pull/1013/files#diff-dd6fab5d1e55a1d860ff8bd2190f145f2574100f734d382cee56c48bd7a7f1f5R43-R49 ?

If I follow your think, it seems we would need a `titiler.multidim` and a `titiler.zarr` packages 🤷 

What if we make the dependencies optional? I'm going to open a PR on top of this one to try some things

src/titiler/xarray/tests/test_factory.py
```diff
@@ -0,0 +1,323 @@
...
+    [0, 1, 2],
+)
+def test_zarr_group(group, app):
+    """Test /tiles endpoints."""
+    resp = app.get(
+        f"/md/tiles/WebMercatorQuad/{group}/0/0.tif",
+        params={"url": zarr_pyramid, "variable": "dataset", "group": group},
+    )
+    assert resp.status_code == 200
+    with MemoryFile(resp.content) as mem:
```
hrodmn, line 314:
```suggestion
    # see src/titiler/xarray/tests/fixtures/generate_fixtures.ipynb
    # for structure of zarr pyramid
    with MemoryFile(resp.content) as mem:
```

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,306 @@
...
+            if da["time"].dtype == "O":
+                da["time"] = da["time"].astype("datetime64[ns]")
+
+            da = da.sel(
+                time=numpy.array(time_as_str, dtype=numpy.datetime64), method="nearest"
+            )
+        else:
+            da = da.isel(time=0)
+
+    assert len(da.dims) in [2, 3], "titiler.xarray can only work with 2D or 3D dataset"
```
hrodmn, line 232:
```suggestion
    if len(da.dims) in [2, 3]:
        raise ValueError("titiler.xarray can only work with 2D or 3D dataset")
```

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,306 @@
...
+    # Make sure we have a valid CRS
+    crs = da.rio.crs or "epsg:4326"
+    da = da.rio.write_crs(crs)
+
+    if crs == "epsg:4326" and (da.x > 180).any():
+        # Adjust the longitude coordinates to the -180 to 180 range
+        da = da.assign_coords(x=(da.x + 180) % 360 - 180)
+
+        # Sort the dataset by the updated longitude coordinates
+        da = da.sortby(da.x)
```
hrodmn, line 216: I wonder if there are more CRS definitions we would want to apply this fix to. Maybe there is a way to tell if we want to adjust coordinates based on some other CRS properties besides an exact name match.

hrodmn (commented):
This is great! The concept of creating pyramids in a zarr store was new to me, then I googled around and found @maxrjones's [notebook](https://developmentseed.org/tile-benchmarking/01-generate-datasets/generate-cmip6-pyramid.html) :laughing:.

It is great to have the `io` methods standardized here so we can import them in `titiler.cmr` and other applications.

src/titiler/xarray/titiler/xarray/io.py
vincentsarago: I guess you want this to be `if not`

src/titiler/xarray/titiler/xarray/io.py
vincentsarago: rasterio doesn't have a `CRS.is_geographic` method yet (https://github.com/rasterio/rasterio/pull/3218) but once it's available we could check if the CRS is geographic and then run those fixes

src/titiler/xarray/titiler/xarray/extensions.py
```diff
@@ -0,0 +1,38 @@
...
+            response_model=List[str],
+            responses={200: {"description": "Return Xarray Dataset variables."}},
+        )
+        def variables(
+            src_path=Depends(factory.path_dependency),
+            io_params=Depends(self.io_dependency),
+        ):
+            """return available variables."""
+            with self.dataset_opener(src_path, **io_params.as_dict()) as ds:
+                return list(ds.data_vars)  # type: ignore
```
vincentsarago, line 38: easier to put this into an extension than to have a `list_variable` in a Class Method (because in the class method the `dataset_opener` is customizable

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,257 @@
...
+    def get(self, key: str) -> bytes:
+        """Get key."""
+        ...
+
+    def set(self, key: str, body: bytes) -> None:
+        """Set key."""
+        ...
+
+
+def xarray_open_dataset(  # noqa: C901
```
vincentsarago, line 27: I've moved everything within the `xarray_open_dataset` function to ease the customization and also because I've made som dependencies optional

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,258 @@
...
+
+@attr.s
+class Reader(XarrayReader):
+    """Reader: Open Zarr file and access DataArray."""
+
+    src_path: str = attr.ib()
+    variable: str = attr.ib()
+
+    # xarray.Dataset options
+    opener: Callable[..., xarray.Dataset] = attr.ib(default=xarray_open_dataset)
```
vincentsarago, line 218: For now the `opener` MUST be a callable that take 4 arguments:
- src_path: str
- group: Any
- decode_times: bool 
- cache_client 

we might change this in the future

hrodmn (approved):
This all looks great, thanks for all of your work moving xarray support further upstream! I left a few small comments but nothing blocking.

src/titiler/xarray/titiler/xarray/dependencies.py
```diff
@@ -0,0 +1,124 @@
...
+    max_size: Annotated[Optional[int], "Maximum image size to read onto."] = None
+    height: Annotated[Optional[int], "Force output image height."] = None
+    width: Annotated[Optional[int], "Force output image width."] = None
+    resampling_method: Annotated[
+        Optional[RIOResampling],
+        Query(
+            alias="resampling",
+            description="RasterIO resampling algorithm. Defaults to `nearest`.",
+        ),
+    ] = None
```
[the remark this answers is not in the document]
vincentsarago: it defaults to what rio-tiler has (`nearest`)

src/titiler/xarray/titiler/xarray/factory.py
```diff
@@ -0,0 +1,212 @@
...
+    reader_dependency: Type[DefaultDependency] = XarrayParams
+
+    # Indexes/Expression Dependencies (Not layer dependencies for Xarray)
+    layer_dependency: Type[DefaultDependency] = DefaultDependency
+
+    # Dataset Options (nodata, reproject)
+    dataset_dependency: Type[DefaultDependency] = DatasetParams
+
+    # Tile/Tilejson/WMTS Dependencies  (Not used in titiler.xarray)
+    tile_dependency: Type[DefaultDependency] = DefaultDependency
```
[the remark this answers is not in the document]
vincentsarago: no because there is not buffer nor padding for the XarrayReader's tile method

vincentsarago:
> But do we need to redefine all of the class attributes (e.g. stats_dependency) that are already defined in titiler.core.factory.TilerFactory?

@hrodmn for most of them we don't have to

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,230 @@
...
+                x for x in ["lat", "latitude", "LAT", "LATITUDE", "Lat"] if x in da.dims
+            )
+            longitude_var_name = next(
+                x
+                for x in ["lon", "longitude", "LON", "LONGITUDE", "Lon"]
+                if x in da.dims
+            )
+        except StopIteration as e:
+            raise ValueError(f"Couldn't find X/Y dimensions in {da.dims}") from e
+
```
j08lue, line 108: Should we add a check here that the `x` and `y` arrays are 1-dimensional?

For example the [UKCP 2.2 km dataset](https://data.ceda.ac.uk/badc/ukcp18/data/land-cpm/uk/2.2km/rcp85/01/tas/mon/v20210615) uses `latitude` and `longitude` as dimension names, but these are 2D fields and reading them with Rasterio/GDAL requires some tweaks https://gist.github.com/j08lue/e792b3c912c33e9191734af7e795b75c

j08lue (commented):
Reviewed the io logic, please see comments below. Nothing blocking.

src/titiler/xarray/titiler/xarray/io.py
j08lue: Good call to keep the heuristics to a minimum.

src/titiler/xarray/pyproject.toml
```diff
@@ -0,0 +1,70 @@
...
+license = {text = "MIT"}
+keywords = [
+    "TiTiler",
+    "Xarray",
+    "Zarr",
+    "NetCDF",
+    "HDF",
+]
+classifiers = [
+    "Intended Audience :: Information Technology",
```
maxrjones, line 19:
```suggestion
    "Development Status :: 4 - Beta",
    "Intended Audience :: Information Technology",
```
I think it would be useful to denote that this is less mature than the other titiler components in the project metadata

vincentsarago:
@maxrjones we've removed the `reference` part but kept the `group` option. There is no `pyramid` support per-say just some tests fixtures which is used to test the `group` option

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -0,0 +1,232 @@
...
+    # Fallback to Zarr
+    else:
+        if protocol == "reference":
+            xr_open_args.update(
+                {
+                    "consolidated": False,
+                    "backend_kwargs": {"consolidated": False},
+                }
+            )
```
vincentsarago, line 80: @maxrjones oh you were talking about this! I think we can let this go 🙏

vincentsarago: I'll create a new issue to add kerchunk reference!

PR #1016, opened 2024-10-31
use xarray.open_zarr and make aiohttp and s3fs optional

vincentsarago:
This PR does:
- remove `reference` and `consolidated` options 
- fall back to `open_zarr` when not dealing with netcdf files
- make `aiohttp` and `s3fs` optional dependencies

Note: 
- we could also make `h5netcdf` and `zarr` optional but for this first version I think it's better to at least have something that support both format 
- in Theory, a user could provider its own `opener` method (`Callable[..., xarray.Dataset]`) making all the custom xarray dataset opener optional


abarciauskas-bgse:
This could still work with kerchunk references, do you know? If not I will try to test it. It may not matter if we go completely with icechunk stores (which should have the same access API as vanilla zarr stores) but it would be good to know if we were losing support for kerchunk references.

vincentsarago:
@abarciauskas-bgse good point, I haven't tested but I will 🙏

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -68,8 +67,14 @@ def get_filesystem(
...
             else s3fs.S3Map(root=src_path, s3=s3_filesystem)
         )
-
-    elif protocol in ["https", "http", "file"]:
+    elif protocol == "reference" or src_path.lower().endswith(".json"):
+        reference_args = {
+            "fo": src_path.replace("reference://", ""),
+            "remote_options": {"anon": anon},
+        }
+        return fsspec.filesystem("reference", **reference_args).get_mapper("")
```
vincentsarago, line 75: check if we have a `reference://` url or a `.json` file

vincentsarago:
@abarciauskas-bgse I've added (and tested) back the Kerchunk reference support but I would love a second pair of 👀 to check 🙏

vincentsarago:
😬 tests are failing because the Kerchunk ref has absolute URL 😬

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -59,18 +59,27 @@ def get_filesystem(
...
         )
-
-    elif protocol == "reference":
-        reference_args = {"fo": src_path, "remote_options": {"anon": anon}}
+    elif protocol == "reference" or src_path.lower().endswith(".json"):
+        reference_args = {
+            "fo": src_path.replace("reference://", ""),
+            "remote_options": {"anon": anon},
+        }
         return fsspec.filesystem("reference", **reference_args).get_mapper("")
```
maxrjones, line 75: there are at least three ways to store virtual references:
- json following kerchunk spec
- parquet following kerchunk spec
- icechunk spec (currently mostly messagepack)

Do you think the "reference" protocol will be used for all of these in the future, with some automatic differentiation between the storage options? Another option would be to specifically have "kerchunk" and "icechunk" as protocol options, with automatic differentiations between json and parquet based on the file extension.

src/titiler/xarray/pyproject.toml
```diff
@@ -30,24 +30,31 @@ classifiers = [
...
     "xarray",
     "rioxarray",
-    "zarr",
     "fsspec",
-    "s3fs",
-    "aiohttp",
-    "pandas",
-    "httpx",
+    "zarr",
+    "h5netcdf",
```
maxrjones, line 37: IMO Zarr and h5netcdf should be optional dependencies so people could deploy titiler-xarray with only one option to keep image size down. This may also make it easier to support other HDF5 readers that work better for remote files in the future (e.g., https://github.com/gauteh/hidefix).
```suggestion
```

src/titiler/xarray/pyproject.toml
```diff
@@ -30,24 +30,31 @@ classifiers = [
...
+s3 = [
+    "s3fs",
+]
+http = [
+    "aiohttp",
+]
+all = [
+    "s3fs",
+    "aiohttp",
+]
```
maxrjones, line 51:
```suggestion
zarr = [
    "zarr",
]
hdf5 = [
    "h5netcdf",
]
all = [
    "s3fs",
    "aiohttp",
    "zarr",
    "hdf5",
]
```
Suggestion to align with above comment about optional dependencies

maxrjones (approved):
nice!

src/titiler/xarray/titiler/xarray/io.py
vincentsarago: I guess we could just support `reference://` prefix instead of checking the `.json` 

I'm also not sure how to support `S3://` referenced files, I'm reading that you can use `"remote_protocol": "s3",` but we don't have ways to know this! 

cc @abarciauskas-bgse

vincentsarago: https://github.com/developmentseed/titiler/pull/1016/commits/8e10da4a2f1defe4bbe363931931ae866993fdeb 

only support `reference` with `reference://` prefix and added support for parquet file (but `fastparquet` dependency should be installed)

src/titiler/xarray/pyproject.toml
vincentsarago: @maxrjones I'll take care of this either in the main PR or in another one later

src/titiler/xarray/titiler/xarray/io.py
abarciauskas-bgse: I'm going to try and figure out how to test this locally but in the mean time, here is what I am wondering: 
* `json following kerchunk spec` - the current implementation should work for local json references (`reference://file.json`) but would it work for remotely stored references, e.g. `s3://file.json` or `http://file.json`?
* `parquet following kerchunk spec` - files will typically end in `.parq` or `.parquet` - again I think this would work for local parquet stores but not sure about remote parquet stores which will be using the s3 or https protocol
* @maxrjones have you tested opening an icechunk store with fsspec and passing that as an argument to xarray? I haven't so I am not sure if it will work, but I think we can address icechunk support later.

If we have `src_path.lower().endswith(".json")` I feel like we should also have `.parquet` conditional here, but I think this is too coupled with what formats kerchunk is implemented in and we should rely on something else... like a path parameter 😬

vincentsarago: in the latest commit I've removed the `.json` test to only accept `reference://` prefix

I guess we could accept things like `reference+s3://` 🤷

abarciauskas-bgse: oh I like `reference+s3://`

maxrjones: > @maxrjones have you tested opening an icechunk store with fsspec and passing that as an argument to xarray? I haven't so I am not sure if it will work, but I think we can address icechunk support later.

no, I've only open icechunk stores without using fsspec

vincentsarago: @abarciauskas-bgse 🤔  I'm not sure we can really support non-local reference file! 

FYI, I've tried `reference://s3://ds-satellite/netcdf/reference.json` and it correctly read the reference file but fails at reading the referenced netcdf (stored on S3 as well). I guess I need to set `remote_protocol="s3"` somewhere. 

@abarciauskas-bgse are we using kerchunk in any production project? to me it seems this need more `input` definition

abarciauskas-bgse: We are using it here: https://www.earthdata.nasa.gov/dashboard/data-catalog/combined_CMIP6_daily_GISS-E2-1-G_tas_kerchunk_DEMO (which I believe uses an s3 protocol for the reference file (private bucket) and the netcdfs it points to are in a public s3 bucket) so I know it should work. I can take a closer look later today ...

abarciauskas-bgse: by the way, I'm on the fence about whether we should continue to support kerchunk-style references, since I think we will be able to use icechunk for reference-style (aka virtual) access in the future. Anyways, kerchunk-style references are still supported by titiler-xarray so we wouldn't have to deprecate https://www.earthdata.nasa.gov/dashboard/data-catalog/combined_CMIP6_daily_GISS-E2-1-G_tas_kerchunk_DEMO right away.

@maxrjones thoughts here?

maxrjones: I can see the motivation for dropping support, since it seems that icechunk is committed to avoiding the fsspec method of using keyword arguments for configuration which is the source of a lot of the complexity here.

abarciauskas-bgse: Ya, ok, @vincentsarago sorry for the delay in decision here, but lets just drop reference support for now and we can add it back if we decide we need it. 

Ideally we could have some documentation about how to easily add it, if people want kerchunk-style reference support with titiler, but I'm not sure at this point how we would do that.

PR #1019, opened 2024-11-05
remove cache layer

vincentsarago:
This PR does:
- update docs
- remove the cache_layer from the default reader but add a notebook to show how to add it back


src/titiler/xarray/README.md
```diff
@@ -41,6 +55,34 @@ titiler/
...
         ├── dependencies.py      - titiler-xarray dependencies
+        ├── extentions.py        - titiler-xarray extensions
         ├── io.py                - titiler-xarray Readers
         └── factory.py           - endpoints factory
 ```
+
+## Custom Dataset Opener
+
+A default Dataset IO is provided within `titiler.xarray.Reader` class but will require optional dependencies (fsspec, zarr, h5netcdf, ...) to be installed with `python -m pip install "titiler.xarray[all]"`.
+The goal of making those dependencies optional is for the user to use their own data access process.
```
abarciauskas-bgse, line 66:
```suggestion
Dependencies are optional so the entire package size can be optimized to only include dependencies required by a given application.
```

src/titiler/xarray/notebooks/xarray_dataset_cache.ipynb
```diff
@@ -0,0 +1,150 @@
+{
```
abarciauskas-bgse, line 1: 👍🏽

PR #1046, opened 2024-12-07
refactor: simplify image format dtype validation

pratapvardhan:
This PR consolidates dtype validation logic by using a lookup. Keeps the existing functionality intact, probably reduces repetitive code and makes it maintainable? 

note: src/titiler/core tests are passing.

cc @vincentsarago – that said, if you feel this change isn’t really necessary, feel free to reject the PR. :)


vincentsarago:
thanks @pratapvardhan 

PR #1048, opened 2024-12-08
cleanup: migrate bootstrap5, remove unused css, fix indent, reuse bs classes

pratapvardhan:
This PR migrates `application/templates/index.html` from bootstrap 4 to bootstrap 5, keeps the visual consistent as before.
- Removes jquery dependency
- Remove unused css
- Reuse bootstrap classes instead of inline-styles.
- Remove stray </div> end tag
- Fix html indentation
cc @vincentsarago 
image


vincentsarago:
thanks @pratapvardhan 

PR #1050, opened 2024-12-10
improve query string handling in LowerCaseQueryStringMiddleware using urlencode

pratapvardhan:
Earlier code manually built the query string by concatenating strings and handling URL encoding one parameter at a time. New code uses list comprehension and built-in urlencode. cc @vincentsarago


pratapvardhan:
cc @vincentsarago would need your approval to run the workflow.

PR #1051, opened 2024-12-10
feature: add layer control to map viewer html template

hrodmn:
- make it possible to toggle the tile layer on/off with a `LayerControl` box
- add ESRI Satellite basemap as an option in addition to OSM
- reformat map.html with prettier


vincentsarago:
thanks @hrodmn 

vincentsarago (approved):

CHANGES.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### titiler.core
+* Add layer control to map viewer template
```
vincentsarago, line 6:
```suggestion
* Add layer control to map viewer template (author @hrodmn, https://github.com/developmentseed/titiler/pull/1051)
```

src/titiler/core/titiler/core/templates/map.html
```diff
@@ -1,146 +1,182 @@
...
-  attributionControl: false
-});
-
-L.control.attribution({prefix: '<a href="https://leafletjs.com" target="_blank">Leaflet</a> | <a href="https://developmentseed.org/titiler" target="_blank">Titiler</a>'}).addTo(map)
-
-const nullIsland = L.marker([0, 0]).addTo(map);
-const madrid = L.marker([40, -3]).addTo(map);
-const london = L.marker([51.50722, -0.1275]).addTo(map)
-const auckland = L.marker([-36.864664, 174.792059]).addTo(map);
-const seattle = L.marker([47.596842, -122.333087]).addTo(map);
```
vincentsarago, line 74: I'm fine removing those marker. I've used them to control the map was correct when using custom TMS. Maybe we should find something better like a really light geojson

hrodmn: Something like a geojson around the Blagden Alley building?

hrodmn: Or the Bermuda Triangle?

vincentsarago: yeah that would be nice!

hrodmn: Actually, I'll just keep the familiar point locations but add a keyboard shortcut (and/or a tiny button) to toggle them on/off.

hrodmn: How does this look?
image

I added a keyboard shortcut (`b`) for toggling the original placemarks on and off, and a tiny button at the bottom right. They are not displayed by default but are easy to toggle on.

PR #1052, opened 2024-12-10
WMTS BoundingBox fix and WMTS template adjustment

AndrewAnnex:
fixes a bug where the WMTS BoundingBox coordinates flipped incorrectly if boundingbox element is used, as the coords must match the CRS order

updates WMTS XML template to avoid placing extra new lines in various places


src/titiler/core/titiler/core/templates/wmts.xml
```diff
@@ -51,13 +51,13 @@
             </TileMatrixSetLink>
             <ResourceURL format="{{ media_type }}" resourceType="tile" template="{{ layer.tiles_url }}?{{ layer.query_string | escape }}" />
         </Layer>
-    {% endfor %}
+    {%- endfor %}
```
vincentsarago, line 54: why do we need `-` ?

vincentsarago:
thanks @AndrewAnnex 

src/titiler/core/titiler/core/templates/wmts.xml
AndrewAnnex: it removes extra white space that's being inserted https://jinja.palletsprojects.com/en/stable/templates/#whitespace-control

AndrewAnnex:
hmm I don't know why the black formatting check is failing, last commit I ran it locally and I just did again via the pre-commit (which I keep forgetting about installing on new systems I'm deving on)

PR #1056, opened 2024-12-18
Include rasterio.show_versions() output in healthz endpoint

scottyhq:
This is a hasty attempt to add `rasterio.show_versions()` to the `healthz` endpoint as discussed in https://github.com/developmentseed/titiler/discussions/731 . It "works" but formatting could be improved...

```bash
curl -X 'GET' \
  'http://127.0.0.1:8000/healthz' \
  -H 'accept: application/json' | jq
```

```
{
  "Hello from titiler!": [
    "rasterio info:",
    "  rasterio: 1.4.3",
    "      GDAL: 3.9.3",
    "      PROJ: 9.4.1",
    "      GEOS: 0.0.0",
    " PROJ DATA: /Users/scotthenderson/miniforge3/envs/titiler/lib/python3.11/site-packages/rasterio/proj_data",
    " GDAL DATA: /Users/scotthenderson/miniforge3/envs/titiler/lib/python3.11/site-packages/rasterio/gdal_data",
    "",
    "System:",
    "    python: 3.11.11 | packaged by conda-forge | (main, Dec  5 2024, 14:21:42) [Clang 18.1.8 ]",
    "executable: /Users/scotthenderson/miniforge3/envs/titiler/bin/python",
    "   machine: macOS-15.2-arm64-arm-64bit",
    "",
    "Python deps:",
    "    affine: 2.4.0",
    "     attrs: 24.3.0",
    "   certifi: 2024.12.14",
    "     click: 8.1.7",
    "     cligj: 0.7.2",
    "    cython: None",
    "     numpy: 2.2.0",
    "click-plugins: None",
    "setuptools: 75.6.0"
  ]
}
```

### DIFF
--- a/src/titiler/application/titiler/application/main.py
+++ b/src/titiler/application/titiler/application/main.py
@@ -1,13 +1,11 @@
 """titiler app."""
-import io
 import logging
 import re
-from contextlib import redirect_stdout
 
 import jinja2
+import rasterio
 from fastapi import Depends, FastAPI, HTTPException, Security
 from fastapi.security.api_key import APIKeyQuery
-from rasterio import show_versions
 from rio_tiler.io import Reader, STACReader
 from starlette.middleware.cors import CORSMiddleware
 from starlette.requests import Request
@@ -217,10 +215,15 @@ if api_settings.lower_case_query_parameters:
 )
 def ping():
     """Health check."""
-    with redirect_stdout(io.StringIO()) as f:
-        show_versions()
-    rasterio_versions = f.getvalue().splitlines()
-    return {"Hello from titiler!": rasterio_versions}
+    return {
+        "versions": {
+            "titiler": titiler_version,
+            "rasterio": rasterio.__version__,
+            "gdal": rasterio.__gdal_version__,
+            "proj": rasterio.__proj_version__,
+            "geos": rasterio.__geos_version__,
+        }
+    }
 
 
 @app.get("/", response_class=HTMLResponse, include_in_schema=False)
--- a/src/titiler/application/titiler/application/main.py
+++ b/src/titiler/application/titiler/application/main.py
@@ -1,11 +1,13 @@
 """titiler app."""
-
+import io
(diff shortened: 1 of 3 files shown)


src/titiler/application/titiler/application/main.py
```diff
@@ -215,7 +217,10 @@ def validate_access_token(access_token: str = Security(api_key_query)):
 )
 def ping():
     """Health check."""
-    return {"ping": "pong!"}
+    with redirect_stdout(io.StringIO()) as f:
+        show_versions()
+    rasterio_versions = f.getvalue().splitlines()
+    return {"Hello from titiler!": rasterio_versions}
```
scottyhq, line 223: This is admittedly trickier than I expected! I was wondering if it would be better to just use rasterio._get_deps_info() even though it's a private method because the python dictionary would be nicer than capturing a multiline string 😅

vincentsarago: 👍 we could maybe simply show 

https://github.com/rasterio/rasterio/blob/64474c1795d5d47c03e271386d21c3b244769946/rasterio/__init__.py#L85-L88

```python
def ping():
     """Health check."""
     return {
          "versions": {
               "titiler": titiler_version,
               "rasterio": rasterio.__version__,
               "gdal": rasterio.__gdal_version__,
               "proj": rasterio.__proj_version__,
               "geos": rasterio.__geos_version__,
          }
     }
```

scottyhq: Agreed that is better! The GEOS version oddly does not seem correct, but I guess that's an upstream issue...

```python
{
  "versions": {
    "titiler": "0.19.2",
    "rasterio": "1.4.3",
    "gdal": "3.9.3",
    "proj": "9.4.1",
    "geos": "0.0.0"
  }
}
```

vincentsarago (approved):

vincentsarago:
thanks @scottyhq 🙏

PR #1060, opened 2024-12-20
TST: use app, client fixture to reuse in tests

pratapvardhan:
@vincentsarago – thoughts on reusing fixtures? Reduces the need to keep defining client in every case.

### DIFF
--- a/src/titiler/core/tests/test_case_middleware.py
+++ b/src/titiler/core/tests/test_case_middleware.py
@@ -17,13 +17,16 @@ def app():
     app.add_middleware(LowerCaseQueryStringMiddleware)
     return app
 
+
 @pytest.fixture
 def client(app):
     """Create test client."""
     return TestClient(app)
 
+
 def test_lowercase_middleware(app, client):
     """Make sure upper and lower case QS are accepted."""
+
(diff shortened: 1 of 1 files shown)


src/titiler/core/tests/test_case_middleware.py
```diff
@@ -2,71 +2,70 @@
...
+    return app
+
+
+@pytest.fixture
+def client(app):
+    """Create test client."""
+    return TestClient(app)
+
+
+def test_lowercase_middleware(app, client):
```
vincentsarago, line 27: We don't need the `app` fixture here then

vincentsarago: ah no we need it, But then it's a bit weird that we register the endpoint after creating the application 😬

pratapvardhan: That's correct, we can leave it as it is and close this PR if the change doesn't seem clean enough.

vincentsarago: Yeah I'm not convinced we need this change, sorry 🙏

PR #1082, opened 2025-01-28
Support Zarr-Python 3

### LINKED ISSUE #1068: [titiler.xarray] update for Zarr 3.0
https://zarr.dev/blog/zarr-python-3-release/

we also need to pin zarr to 2.0 https://github.com/developmentseed/titiler/actions/runs/12695652933/job/35388000642?pr=1067

maxrjones:
Adds support for both opening Zarr stores using either Zarr-Python 2 or Zarr-Python 3.

Fixed #1068

### DIFF
--- a/src/titiler/xarray/pyproject.toml
+++ b/src/titiler/xarray/pyproject.toml
@@ -68,15 +68,6 @@ test = [
     "h5netcdf",
     "fsspec",
 ]
-test-zarr-v2 = [
-    "pytest",
-    "pytest-cov",
-    "pytest-asyncio",
-    "httpx",
-    "zarr<3.0",
-    "h5netcdf",
-    "fsspec",
-]
 
 [project.urls]
 Homepage = "https://developmentseed.org/titiler/"
--- a/src/titiler/xarray/pyproject.toml
+++ b/src/titiler/xarray/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
 
 [project.optional-dependencies]
 full = [
-    "zarr>=2,<3",
+    "zarr",
     "h5netcdf",
     "fsspec",
     "s3fs",
@@ -46,7 +46,7 @@ full = [
     "gcsfs",
 ]
 minimal = [
-    "zarr>=2,<3",
+    "zarr",
     "h5netcdf",
     "fsspec",
 ]
@@ -64,7 +64,16 @@ test = [
     "pytest-cov",
     "pytest-asyncio",
     "httpx",
-    "zarr>=2,<3",
+    "zarr",
+    "h5netcdf",
+    "fsspec",
+]
+test-zarr-v2 = [
+    "pytest",
+    "pytest-cov",
+    "pytest-asyncio",
+    "httpx",
+    "zarr<3.0",
     "h5netcdf",
     "fsspec",
 ]
--- a/src/titiler/xarray/titiler/xarray/io.py
+++ b/src/titiler/xarray/titiler/xarray/io.py
@@ -53,7 +53,6 @@ def xarray_open_dataset(  # noqa: C901
     else:
         assert zarr is not None, "'zarr' must be installed to read Zarr dataset"
         xr_engine = "zarr"
-        _zarr_v3 = module_available("zarr", minversion="3.0")
 
     # Arguments for xarray.open_dataset
     # Default args
--- a/src/titiler/xarray/titiler/xarray/io.py
+++ b/src/titiler/xarray/titiler/xarray/io.py
@@ -9,6 +9,7 @@ import xarray
 from morecantile import TileMatrixSet
 from rio_tiler.constants import WEB_MERCATOR_TMS
 from rio_tiler.io.xarray import XarrayReader
+from xarray.namedarray.utils import module_available
 
 
 def xarray_open_dataset(  # noqa: C901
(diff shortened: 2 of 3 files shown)


src/titiler/xarray/pyproject.toml
```diff
@@ -64,7 +64,16 @@ test = [
...
+]
+test-zarr-v2 = [
+    "pytest",
+    "pytest-cov",
+    "pytest-asyncio",
+    "httpx",
+    "zarr<3.0",
     "h5netcdf",
     "fsspec",
 ]
```
maxrjones, line 79: This is just for testing. I'm more familiar with hatch than pdm, but if we want more comprehensive tests for both minimum versions and/or upstream main branches I could work on setting that up.

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -66,55 +52,8 @@ def xarray_open_dataset(  # noqa: C901
...
-
-        filesystem = fsspec.filesystem(protocol)  # type: ignore
-        file_handler = (
-            filesystem.open(src_path)
-            if xr_engine == "h5netcdf"
-            else filesystem.get_mapper(src_path)
-        )
-
-    else:
-        raise ValueError(f"Unsupported protocol: {protocol}, for {src_path}")
```
maxrjones, line 117: It's a lot simpler to rely on fsspec's import error message and use fsspec.filesystem(protocol). This also provides the benefit of supporting a wider range of protocols. Still, let me know if you'd rather control the error message through the removed approach.

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -135,13 +74,21 @@ def xarray_open_dataset(  # noqa: C901
...
+        if _zarr_v3:
+            if protocol == "file":
+                store = zarr.storage.LocalStore(parsed.path, read_only=True)
+            else:
+                fs = fsspec.filesystem(protocol, storage_options={"asynchronous": True})
+                store = zarr.storage.FsspecStore(fs, path=src_path, read_only=True)
+            ds = xarray.open_zarr(store, **xr_open_args)
+        else:
+            fs = fsspec.filesystem(protocol)
+            ds = xarray.open_zarr(fs.get_mapper(src_path), **xr_open_args)
```
maxrjones, line 91: This is the bulk of the changes to account for Zarr no longer accepting fsspec's mutable mapping wrappers and expecting asynchronous file systems.

src/titiler/xarray/titiler/xarray/io.py
vincentsarago: I think @abarciauskas-bgse wrote most of this originally in titiler-xarray

src/titiler/xarray/pyproject.toml
```diff
@@ -64,7 +64,16 @@ test = [
...
+    "zarr",
+    "h5netcdf",
+    "fsspec",
+]
+test-zarr-v2 = [
+    "pytest",
+    "pytest-cov",
+    "pytest-asyncio",
+    "httpx",
+    "zarr<3.0",
```
vincentsarago, line 76: I would be fine only testing with zarr>3.0 TBH

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -66,55 +52,8 @@ def xarray_open_dataset(  # noqa: C901
...
-        filesystem = fsspec.filesystem(protocol)  # type: ignore
-        file_handler = (
-            filesystem.open(src_path)
-            if xr_engine == "h5netcdf"
-            else filesystem.get_mapper(src_path)
-        )
-
-    else:
-        raise ValueError(f"Unsupported protocol: {protocol}, for {src_path}")
+        _zarr_v3 = module_available("zarr", minversion="3.0")
```
vincentsarago, line 56: can we move this a L82?

src/titiler/xarray/titiler/xarray/io.py
vincentsarago, line 91:
```suggestion
        if module_available("zarr", minversion="3.0"):
            if protocol == "file":
                store = zarr.storage.LocalStore(parsed.path, read_only=True)
            else:
                fs = fsspec.filesystem(protocol, storage_options={"asynchronous": True})
                store = zarr.storage.FsspecStore(fs, path=src_path, read_only=True)
            
        else:
            store = fsspec.filesystem(protocol).get_mapper(src_path)

        ds = xarray.open_zarr(store, **xr_open_args)
```

src/titiler/xarray/pyproject.toml
```diff
@@ -38,15 +38,15 @@ dependencies = [
...
+    "zarr",
     "h5netcdf",
     "fsspec",
     "s3fs",
     "aiohttp",
     "gcsfs",
 ]
 minimal = [
-    "zarr>=2,<3",
+    "zarr",
```
abarciauskas-bgse, line 49: will this work for zarr<3 ? Should that be specified?

src/titiler/xarray/titiler/xarray/io.py
abarciauskas-bgse: This approach seems better to me too @maxrjones  I didn't know about fsspec's import error message and `fsspec.filesystem(protocol)`

src/titiler/xarray/pyproject.toml
maxrjones: Yes, I accounted for both the new and old syntax in https://github.com/developmentseed/titiler/blob/e081d0f1452e4a950d8e9c13c939f0c5627956aa/src/titiler/xarray/titiler/xarray/io.py#L81-L89 and tested locally for zarr<3.

PR #1089, opened 2025-01-29
Add output min and max metadata to slope algorithm

tayden:
This is just a small tweak to add more output metadata to the slope algorithm.


vincentsarago:
Would you mind adding a note in the changelog 🙏

PR #1098, opened 2025-02-24
Update Dockerfile to move handler and set command for Lambda function

wakame1367:
## What I am changing
- Moving all files from the `/asset` directory into the Lambda task root (`${LAMBDA_TASK_ROOT}`).
- Updating the CMD instruction to invoke the correct Lambda handler (`handler.handler`) instead of the placeholder echo command.

## How I did it
- Inserted the command `RUN mv /asset/* ${LAMBDA_TASK_ROOT}/` in the Dockerfile to move all asset files to the correct directory.
- Changed the CMD line from `CMD ["echo", "hello world"]` to `CMD ["handler.handler"]`, ensuring AWS Lambda calls the proper function defined in `handler.py`.

## How you can test it
- Build the Docker image locally using the updated Dockerfile.
- Run the image with the Lambda runtime interface emulator (or deploy to AWS Lambda) and invoke the function.
- Verify that the Lambda function executes the handler from `handler.py` and returns the expected result.

## Related Issues
- (None)

### DIFF
--- a/deployment/aws/lambda/Dockerfile
+++ b/deployment/aws/lambda/Dockerfile
@@ -21,5 +21,5 @@ RUN rm -rdf /asset/numpy/doc/ /asset/boto3* /asset/botocore* /asset/bin /asset/g
 RUN yum remove -y gcc-c++
 
 COPY lambda/handler.py /asset/handler.py
-
-CMD ["echo", "hello world"]
+RUN mv /asset/* ${LAMBDA_TASK_ROOT}/
+CMD ["handler.handler"]


vincentsarago:
Hi @wakame1367 
Thanks for this PR. I'm not sure to get why those changes are needed? 

Do you aim to use the docker image directly in AWS Lambda? We used the docker container just to build the lambda package (for the CDK app).

wakame1367:
@vincentsarago 
Thank you for your feedback. You're absolutely right - I misunderstood the architecture.

I now understand that in this project, the Dockerfile is solely used to build the Lambda package for the CDK application and not intended to be deployed directly as a Lambda container image.

Initially, I was thinking of the approach where Docker images are deployed directly to AWS Lambda, but it's clear now that this repository is designed to deploy through CDK.

I apologize for suggesting unnecessary changes. I'll close this PR.

PR #1106, opened 2025-03-24
Bump cdk to 2.185.0

f-skold:
## Bump cdk to 2.185.0

(Not sure how to use this template)

+ Bump cdk to 2.185.0
+ Use python 3.13 in dockerfiles/Dockerfile

I guess that it should be possible to test this, and when accept it.
I have tested the dockerfile using tiss env:

    docker-compose up --build titiler


vincentsarago:
done in https://github.com/developmentseed/titiler/commit/23af6015a1d6d86fa4c2ea7c4d74616fe202f039

note: still on python 3.12

PR #1115, opened 2025-03-31
Feature/refactor middlewares

### LINKED ISSUE #1107: update logging middleware with more request details
We could include more info about the route, path parameters, query parameters, etc like we recently added to the maap-eoapi deployment in https://github.com/MAAP-Project/maap-eoapi/pull/49

https://github.com/developmentseed/titiler/blob/6d1519e9a289a1f1f8e15c04020920f82cf02c60/src/titiler/core/titiler/core/middleware.py#L100-L135

vincentsarago:
closes #1107 

This PR does:
- refactor middleware classes to use python's dataclasses (a.la @alukach 🙏)
- re-write the LoggerMiddleware to enable more output data and custom logger

### DIFF
--- a/src/titiler/application/titiler/application/main.py
+++ b/src/titiler/application/titiler/application/main.py
@@ -225,6 +225,7 @@ if api_settings.debug:
                                     "method",
                                     "referer",
                                     "origin",
+                                    "route",
                                     "path",
                                     "path_params",
                                     "query_params",
--- a/src/titiler/application/titiler/application/main.py
+++ b/src/titiler/application/titiler/application/main.py
@@ -1,7 +1,9 @@
 """titiler app."""
 
+import json
 import logging
 import re
+from logging import config as log_config
 
 import jinja2
 import rasterio
@@ -204,6 +206,64 @@ app.add_middleware(
 if api_settings.debug:
     app.add_middleware(LoggerMiddleware)
     app.add_middleware(TotalTimeMiddleware)
+
+    log_config.dictConfig(
+        {
+            "version": 1,
+            "disable_existing_loggers": False,
+            "formatters": {
+                "detailed": {
+                    "format": "%(asctime)s - %(levelname)s - %(name)s - %(message)s"
+                },
+                "request": {
+                    "format": (
+                        "%(asctime)s - %(levelname)s - %(name)s - %(message)s "
+                        + json.dumps(
+                            {
+                                k: f"%({k})s"
+                                for k in [
+                                    "method",
+                                    "referer",
+                                    "origin",
+                                    "path",
+                                    "path_params",
+                                    "query_params",
+                                    "headers",
+                                ]
+                            }
+                        )
+                    ),
+                },
+            },
+            "handlers": {
+                "console_detailed": {
+                    "class": "logging.StreamHandler",
+                    "level": "WARNING",
+                    "formatter": "detailed",
+                    "stream": "ext://sys.stdout",
+                },
+                "console_request": {
+                    "class": "logging.StreamHandler",
+                    "level": "DEBUG",
+                    "formatter": "request",
+                    "stream": "ext://sys.stdout",
+                },
+            },
+            "loggers": {
+                "titlier": {
+                    "level": "WARNING",
+                    "handlers": ["console_detailed"],
+                    "propagate": False,
+                },
+                "titiler-requests": {
+                    "level": "INFO",
+                    "handlers": ["console_request"],
+                    "propagate": False,
+                },
+            },
+        }
+    )
+
 
 if api_settings.lower_case_query_parameters:
     app.add_middleware(LowerCaseQueryStringMiddleware)
--- a/src/titiler/application/titiler/application/main.py
+++ b/src/titiler/application/titiler/application/main.py
@@ -202,7 +202,7 @@ app.add_middleware(
 )
 
 if api_settings.debug:
-    app.add_middleware(LoggerMiddleware, headers=True, querystrings=True)
+    app.add_middleware(LoggerMiddleware)
     app.add_middleware(TotalTimeMiddleware)
 
 if api_settings.lower_case_query_parameters:
--- a/src/titiler/core/titiler/core/middleware.py
+++ b/src/titiler/core/titiler/core/middleware.py
@@ -2,13 +2,11 @@
 
 from __future__ import annotations
 
-import json
 import logging
 import re
-import sys
 import time
 from dataclasses import dataclass, field
-from typing import List, Literal, Set, get_args
+from typing import Set
 from urllib.parse import urlencode
 
 from starlette.datastructures import MutableHeaders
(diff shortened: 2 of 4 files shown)


hrodmn (approved):
This looks great, I left one suggestion for extracting the `route` from the request but if you want to skip it that's fine.

src/titiler/core/titiler/core/middleware.py
```diff
@@ -97,57 +95,62 @@ async def send_wrapper(message: Message):
...
-            app (ASGIApp): starlette/FastAPI application.
+def create_logger() -> logging.Logger:
+    """Create titiler logger."""
+    logger = logging.getLogger("titiler")
+    logger.setLevel(logging.INFO)
+    handler = logging.StreamHandler(sys.stdout)
+    handler.setLevel(logging.INFO)
+    handler.setFormatter(logging.Formatter("%(name)s - %(message)s"))
+    logger.addHandler(handler)
+    return logger
```
alukach, line 111: I'm a bit opinionated about logging, so take this with a grain of salt:

1. Any reason that we are hardcoding the logger name? The idiomatic way would be to use `__name__` (which would resolve to `titiler.core.middleware` in this situation). This will allow log handlers to more precisely target logging policies if desired (more on this below).
2. It feels a bit odd to be defining the handler & formatter in a manner so tightly coupled with the logger itself. I understand that this is the _default_ and can be easily overridden, but it seems surprising given that we already have log-handling logic within `main.py`

I think the most flexible systems look like this:

* codebase contains extensivive logging at various log levels via `logging.getLogger(__name__)`
* at the entrypoint of the application (eg `main.py`, `__main__.py`), the loggers are configured to specify the formats, handlers, and log levels for each logger.

src/titiler/core/titiler/core/middleware.py
```diff
@@ -97,57 +95,62 @@ async def send_wrapper(message: Message):
...
+            }
+            data = {k: v for k, v in data.items() if k in self.log_data}
 
-            qs = dict(request.query_params)
-            if qs and self.querystrings:
-                self.logger.debug(qs)
-
-            if self.headers:
-                self.logger.debug(dict(request.headers))
+            self.logger.info(f"Request: {json.dumps(data)}")
```
alukach, line 141: Kind of following above, if we want to allow implementors to specify the log format, they could do that with standard log utils and the [`extra` parameter](https://docs.python.org/3/howto/logging-cookbook.html#adding-contextual-information-to-your-logging-output).  It's a bit funky though...

e.g. 

```py
# in titiler.core.middleware
logger = logging.getLogger('.'.join(__name__, 'request')
logger.info(
    f"Request received: {request.path} {request.method}",
    extra={
        "method": request.method,
        "request": request,
        "referer": next(
            (request.headers.get(attr) for attr in ["referer", "referrer"]), None
        ),
        "origin": request.headers.get("origin"),
        "path": request.path,
        "path_params": request.path_params,
        "query_params": dict(request.query_params),
        "headers": dict(request.headers),
    },
)

# entrypoint
logging.config.dictConfig(
    {
        "version": 1,
        "disable_existing_loggers": False,
        "formatters": {
            "detailed": {
                "format": "%(asctime)s - %(levelname)s - %(name)s - %(message)s"
            },
            "request": {
                "format": (
                    # Their custom format here...
                    f"%(asctime)s - %(levelname)s - %(name)s - %(message)s "
                    + json.dumps(
                        {
                            k: f"%({k})s"
                            for k in [
                                "method",
                                "referer",
                                "origin",
                                "path",
                                "path_params",
                                "query_params",
                                "headers",
                            ]
                        }
                    )
                ),
            },
        },
        "handlers": {
            "console_detailed": {
                "class": "logging.StreamHandler",
                "level": "WARNING",
                "formatter": "detailed",
                "stream": "ext://sys.stdout",
            },
            "console_request": {
                "class": "logging.StreamHandler",
                "level": "DEBUG",
                "formatter": "request",
                "stream": "ext://sys.stdout",
            },
        },
        "loggers": {
            "titlier": {
                "level": "WARNING",
                "handlers": ["console_detailed"],
                "propagate": False,
            },
            "titlier.core.middleware.request": {
                "level": "INFO",
                "handlers": ["console_request"],
                "propagate": False,
            },
        },
    }
)
```

alukach (approved):
I do ❤️ dataclasses

I left some 🌶️ opinions on logging structure, which admittedly isn't a big deal and you can take it or leave it.

src/titiler/core/titiler/core/middleware.py
geospatial-jeff: `extra` is definitely preferred over dumping the dictionary into the log message because `extra` will be indexed by log shippers which makes it searchable in your standard ELK stack, GCP Cloud Logging, AWS CloudWatch etc.

src/titiler/core/titiler/core/middleware.py
```diff
@@ -97,57 +95,62 @@ async def send_wrapper(message: Message):
...
+            data = {
+                "method": request.method,
+                "referer": request.headers.get("referer")
+                or request.headers.get("referrer"),
+                "origin": request.headers.get("origin"),
+                "path": request.url.path,
+                "path_params": request.path_params,
+                "query_params": dict(request.query_params),
+                "headers": dict(request.headers),
+            }
```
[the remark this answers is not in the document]
vincentsarago: I wanted to make things `simple`. Looking for the `route` will be a really `niche` need IMO. 

What's your use case for needing it?

src/titiler/core/titiler/core/middleware.py
```diff
@@ -95,30 +93,13 @@ async def send_wrapper(message: Message):
...
 @dataclass(frozen=True)
 class LoggerMiddleware:
     """MiddleWare to add logging."""
 
     app: ASGIApp
-    logger: logging.Logger = field(default_factory=create_logger)
-    log_data: List[AVAILABLE_LOG_DATA] = field(
-        default_factory=lambda: list(get_args(AVAILABLE_LOG_DATA))
+    logger: logging.Logger = field(
+        default_factory=lambda: logging.getLogger("titiler-requests")
```
vincentsarago, line 102: I choose to set the `name` directly to be a be more precise than `titiler.core.middleware`

src/titiler/application/titiler/application/main.py
```diff
@@ -205,6 +207,64 @@ def validate_access_token(access_token: str = Security(api_key_query)):
...
+                    "propagate": False,
+                },
+                "titiler-requests": {
+                    "level": "INFO",
+                    "handlers": ["console_request"],
+                    "propagate": False,
+                },
+            },
+        }
+    )
```
vincentsarago, line 265: @alukach does this match what you envisioned ?

src/titiler/core/titiler/core/middleware.py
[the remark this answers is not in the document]
vincentsarago: but is `path` already not enough?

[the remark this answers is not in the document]
hrodmn: The use-case I have in mind is just getting a summary of the top `n` requested routes without path or query parameters. You might be able to get that by parsing the `path` values in something like CloudWatch, but you probably will not have access to the application's registered routes in that context so it would be tricky to get it all right.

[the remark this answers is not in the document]
hrodmn: For example, in MAAP we have this CloudWatch dashboard that shows the number of requests by route:
image
It should be helpful for prioritizing improvements in our application based on observed user behavior!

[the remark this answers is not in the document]
vincentsarago: ahhhh you want the original route path 😓 sorry I didn't caught this! 

I think it's possible to get this directly from the request object

hrodmn (approved):

src/titiler/core/titiler/core/middleware.py
alukach: I think it would be wise to at least put it behind `titiler` (ie `titiler.requests`) that way a person could apply a blanket logging policy to all of `titiler` or one purely targeting the requests logger. As currently written, implementors wouldn't be able to target titiler & the titlier-requests with a single statement.

src/titiler/application/titiler/application/main.py
alukach: Yes, aside from the details I put in this comment: https://github.com/developmentseed/titiler/pull/1115/files#r2022394747

src/titiler/core/titiler/core/middleware.py
vincentsarago: I'm not sure to get what you're saying. would you prefer `titiler.requests` ?

alukach: yes, the idea being that if they shared the same root logger (ie `titiler`), someone could customize the logger for `titiler` (which would include all of the `titiler.requests` logs) or they could do separate loggers for `titiler` and `titiler.requests` if they wanted that (setting `propogate=False` to keep the `titiler.requests` logs from going to the `titiler` log handler).

PR #1116, opened 2025-04-01
Fix TerrainRGB algorithm and param user-controlled nodata-height

jo-chemla:
- [x] Fixes Terrain-RGB algorithm 
- [x] Added two params `use_nodata_height and nodata_height` (or could have used a field which could be either undefined or that height value?). 
- [x] Also added user-controlled nodata-height parametrization for terrarium

Can be tested via 
        
> 'tiles': ['http://localhost:8000/cog/tiles/WebMercatorQuad/{z}/{x}/{y}@1x.png?algorithm=terrainrgb&algorithm_params={"use_nodata_height":true,"nodata_height":0.0}&nodata=-9999&url=https://3d.iconem.com/greece/amphipolis/Amphipolis_general_dsm_10cm_cog_cropped_alpha.tif'] 

**TODO**: Also should take a moment to add suggested updates from discussion https://github.com/developmentseed/titiler/discussions/1110
 - [ ] terrarium/terrain-rgb intepretation for dem-related algorithms like hillshade & slope
 - [x] z-exaggeration algorithm-param
 - [x] algorithms doc update
 - [x] ~Just noticed also the /cog/viewer endpoint to help users play with url-params, bands expressions etc. Extend single-band user-parametrization to also allow for algorithms and a few other url-params like rescale etc?~ other PR https://github.com/developmentseed/titiler/pull/1119

### DIFF
--- a/src/titiler/core/titiler/core/algorithm/dem.py
+++ b/src/titiler/core/titiler/core/algorithm/dem.py
@@ -167,8 +167,7 @@ class Terrarium(BaseAlgorithm):
 
     title: str = "Terrarium"
     description: str = "Encode DEM into RGB (Mapzen Terrarium)."
-    use_nodata_height: bool = Field(False)
-    nodata_height: float = Field(0.0, ge=-99999.0, le=99999.0)
+    nodata_height: Optional[float] = Field(None, ge=-99999.0, le=99999.0)
 
     # metadata
     input_nbands: int = 1
--- a/src/titiler/core/titiler/core/algorithm/dem.py
+++ b/src/titiler/core/titiler/core/algorithm/dem.py
@@ -1,6 +1,7 @@
 """titiler.core.algorithm DEM."""
 
 import numpy
+from typing import Optional
 from pydantic import Field
 from rasterio import windows
 from rio_tiler.colormap import apply_cmap, cmap
--- a/src/titiler/core/titiler/core/algorithm/dem.py
+++ b/src/titiler/core/titiler/core/algorithm/dem.py
@@ -235,7 +235,7 @@ class TerrainRGB(BaseAlgorithm):
         if _range_check(datarange):
             raise ValueError(f"Data of {datarange} larger than 256 ** 3")
 
-        if self.use_nodata_height: 
+        if self.nodata_height is not None: 
             data[img.array.mask[0]] = (self.nodata_height - self.baseval) / self.interval
 
         data_int32 = data.astype(numpy.int32)
--- a/src/titiler/core/titiler/core/algorithm/dem.py
+++ b/src/titiler/core/titiler/core/algorithm/dem.py
@@ -22,6 +22,7 @@ class HillShade(BaseAlgorithm):
     azimuth: int = Field(45, ge=0, le=360)
     angle_altitude: float = Field(45.0, ge=-90.0, le=90.0)
     buffer: int = Field(3, ge=0, le=99)
+    z_exaggeration: float = Field(1., ge=1e-6, le=1e6)
 
     # metadata
     input_nbands: int = 1
@@ -31,6 +32,8 @@ class HillShade(BaseAlgorithm):
     def __call__(self, img: ImageData) -> ImageData:
         """Create hillshade from DEM dataset."""
         x, y = numpy.gradient(img.array[0])
+        x *= self.z_exaggeration
+        y *= self.z_exaggeration
         slope = numpy.pi / 2.0 - numpy.arctan(numpy.sqrt(x * x + y * y))
         aspect = numpy.arctan2(-x, y)
         azimuth = 360.0 - self.azimuth
@@ -71,6 +74,7 @@ class Slope(BaseAlgorithm):
 
     # parameters
     buffer: int = Field(3, ge=0, le=99, description="Buffer size for edge effects")
+    z_exaggeration: float = Field(1., ge=1e-6, le=1e6)
 
     # metadata
     input_nbands: int = 1
@@ -86,6 +90,8 @@ class Slope(BaseAlgorithm):
         pixel_size_y = abs(img.transform[4])
 
         x, y = numpy.gradient(img.array[0])
+        x *= self.z_exaggeration
+        y *= self.z_exaggeration
         dx = x / pixel_size_x
         dy = y / pixel_size_y
 
--- a/src/titiler/core/titiler/core/algorithm/dem.py
+++ b/src/titiler/core/titiler/core/algorithm/dem.py
@@ -1,7 +1,8 @@
 """titiler.core.algorithm DEM."""
 
-import numpy
 from typing import Optional
+
+import numpy
 from pydantic import Field
 from rasterio import windows
 from rio_tiler.colormap import apply_cmap, cmap
@@ -23,7 +24,7 @@ class HillShade(BaseAlgorithm):
     azimuth: int = Field(45, ge=0, le=360)
     angle_altitude: float = Field(45.0, ge=-90.0, le=90.0)
     buffer: int = Field(3, ge=0, le=99)
-    z_exaggeration: float = Field(1., ge=1e-6, le=1e6)
+    z_exaggeration: float = Field(1.0, ge=1e-6, le=1e6)
 
     # metadata
     input_nbands: int = 1
@@ -75,7 +76,7 @@ class Slope(BaseAlgorithm):
 
     # parameters
     buffer: int = Field(3, ge=0, le=99, description="Buffer size for edge effects")
-    z_exaggeration: float = Field(1., ge=1e-6, le=1e6)
+    z_exaggeration: float = Field(1.0, ge=1e-6, le=1e6)
 
     # metadata
     input_nbands: int = 1
@@ -178,8 +179,10 @@ class Terrarium(BaseAlgorithm):
     def __call__(self, img: ImageData) -> ImageData:
         """Encode DEM into RGB."""
         data = numpy.clip(img.array[0] + 32768.0, 0.0, 65535.0)
-        if self.nodata_height is not None: 
-            data[img.array.mask[0]] = numpy.clip(self.nodata_height + 32768.0, 0.0, 65535.0)
+        if self.nodata_height is not None:
+            data[img.array.mask[0]] = numpy.clip(
+                self.nodata_height + 32768.0, 0.0, 65535.0
+            )
         r = data / 256
         g = data % 256
         b = (data * 256) % 256
@@ -235,13 +238,15 @@ class TerrainRGB(BaseAlgorithm):
         if _range_check(datarange):
             raise ValueError(f"Data of {datarange} larger than 256 ** 3")
 
-        if self.nodata_height is not None: 
-            data[img.array.mask[0]] = (self.nodata_height - self.baseval) / self.interval
+        if self.nodata_height is not None:
+            data[img.array.mask[0]] = (
+                self.nodata_height - self.baseval
+            ) / self.interval
 
         data_int32 = data.astype(numpy.int32)
-        b = (data_int32) & 0xff 
-        g = (data_int32 >> 8) & 0xff
-        r = (data_int32 >> 16) & 0xff
+        b = (data_int32) & 0xFF
+        g = (data_int32 >> 8) & 0xFF
+        r = (data_int32 >> 16) & 0xFF
 
         return ImageData(
             numpy.ma.stack([r, g, b]).astype(self.output_dtype),
--- a/src/titiler/core/titiler/core/algorithm/dem.py
+++ b/src/titiler/core/titiler/core/algorithm/dem.py
@@ -177,7 +177,7 @@ class Terrarium(BaseAlgorithm):
     def __call__(self, img: ImageData) -> ImageData:
         """Encode DEM into RGB."""
         data = numpy.clip(img.array[0] + 32768.0, 0.0, 65535.0)
-        if self.use_nodata_height: 
+        if self.nodata_height is not None: 
             data[img.array.mask[0]] = numpy.clip(self.nodata_height + 32768.0, 0.0, 65535.0)
         r = data / 256
         g = data % 256
--- a/src/titiler/core/titiler/core/algorithm/dem.py
+++ b/src/titiler/core/titiler/core/algorithm/dem.py
@@ -227,7 +227,7 @@ class TerrainRGB(BaseAlgorithm):
             raise ValueError(f"Data of {datarange} larger than 256 ** 3")
 
         if self.use_nodata_height: 
-            data[img.array.mask[0]] = (0 - self.baseval) / self.interval
+            data[img.array.mask[0]] = (self.nodata_height - self.baseval) / self.interval
 
         data_int32 = data.astype(numpy.int32)
         b = (data_int32) & 0xff 
--- a/src/titiler/core/titiler/core/algorithm/dem.py
+++ b/src/titiler/core/titiler/core/algorithm/dem.py
@@ -191,6 +191,8 @@ class TerrainRGB(BaseAlgorithm):
     # parameters
     interval: float = Field(0.1, ge=0.0, le=1.0)
     baseval: float = Field(-10000.0, ge=-99999.0, le=99999.0)
+    use_nodata_height: bool = Field(False)
+    nodata_height: float = Field(0.0, ge=-99999.0, le=99999.0)
 
     # metadata
     input_nbands: int = 1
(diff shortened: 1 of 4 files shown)


src/titiler/core/titiler/core/algorithm/dem.py
```diff
@@ -224,9 +236,13 @@ def _range_check(datarange):
...
-        r = ((((data // 256) // 256) / 256) - (((data // 256) // 256) // 256)) * 256
-        g = (((data // 256) / 256) - ((data // 256) // 256)) * 256
-        b = ((data / 256) - (data // 256)) * 256
+        if self.use_nodata_height: 
+            data[img.array.mask[0]] = (self.nodata_height - self.baseval) / self.interval
+
+        data_int32 = data.astype(numpy.int32)
+        b = (data_int32) & 0xff 
+        g = (data_int32 >> 8) & 0xff
+        r = (data_int32 >> 16) & 0xff
```
vincentsarago, line 243: is that equivalent? or it does something new?

src/titiler/core/titiler/core/algorithm/dem.py
```diff
@@ -224,9 +236,13 @@ def _range_check(datarange):
         if _range_check(datarange):
             raise ValueError(f"Data of {datarange} larger than 256 ** 3")
 
-        r = ((((data // 256) // 256) / 256) - (((data // 256) // 256) // 256)) * 256
-        g = (((data // 256) / 256) - ((data // 256) // 256)) * 256
-        b = ((data / 256) - (data // 256)) * 256
+        if self.use_nodata_height: 
+            data[img.array.mask[0]] = (self.nodata_height - self.baseval) / self.interval
```
vincentsarago, line 239: I don't think we need the 2 options, just `nodata_height` would be enough

src/titiler/core/titiler/core/algorithm/dem.py
```diff
@@ -191,6 +201,8 @@ class TerrainRGB(BaseAlgorithm):
     # parameters
     interval: float = Field(0.1, ge=0.0, le=1.0)
     baseval: float = Field(-10000.0, ge=-99999.0, le=99999.0)
+    use_nodata_height: bool = Field(False)
+    nodata_height: float = Field(0.0, ge=-99999.0, le=99999.0)
```
vincentsarago:
```suggestion
    nodata_height: Optional[float] = Field(None, ge=-99999.0, le=99999.0)
```

src/titiler/core/titiler/core/algorithm/dem.py
```diff
@@ -224,9 +236,13 @@ def _range_check(datarange):
         if _range_check(datarange):
             raise ValueError(f"Data of {datarange} larger than 256 ** 3")
 
-        r = ((((data // 256) // 256) / 256) - (((data // 256) // 256) // 256)) * 256
-        g = (((data // 256) / 256) - ((data // 256) // 256)) * 256
-        b = ((data / 256) - (data // 256)) * 256
+        if self.use_nodata_height:
```
vincentsarago:
```suggestion
        if self.nodata_height is not None: 
```

src/titiler/core/titiler/core/algorithm/dem.py
jo-chemla: Thanks for the Optional, exactly what I was looking for!

src/titiler/core/titiler/core/algorithm/dem.py
```diff
@@ -170,6 +178,8 @@ class Terrarium(BaseAlgorithm):
     def __call__(self, img: ImageData) -> ImageData:
         """Encode DEM into RGB."""
         data = numpy.clip(img.array[0] + 32768.0, 0.0, 65535.0)
+        if self.use_nodata_height:
```
vincentsarago:
```suggestion
        if self.nodata_height is not None: 
```

src/titiler/core/titiler/core/algorithm/dem.py
```diff
@@ -161,6 +167,8 @@ class Terrarium(BaseAlgorithm):
 
     title: str = "Terrarium"
     description: str = "Encode DEM into RGB (Mapzen Terrarium)."
+    use_nodata_height: bool = Field(False)
+    nodata_height: float = Field(0.0, ge=-99999.0, le=99999.0)
```
vincentsarago:
```suggestion
    nodata_height: Optional[float] = Field(None, ge=-99999.0, le=99999.0)
```

src/titiler/core/titiler/core/algorithm/dem.py
jo-chemla: It is doing things differently, qgis could not interpret the terrain-rgb endpoint as-is previously, but this one does work great. Maybe have to do with floating-point precision errors with that large 256^3 scaling?

src/titiler/core/titiler/core/algorithm/dem.py
```diff
@@ -161,6 +168,7 @@ class Terrarium(BaseAlgorithm):
 
     title: str = "Terrarium"
     description: str = "Encode DEM into RGB (Mapzen Terrarium)."
+    nodata_height: Optional[float] = Field(None, ge=-99999.0, le=99999.0)
```
vincentsarago, line 171: The issue is when you don't know much about the data (e.g no idea about the `min`) 

we could automagically set the `nodata_height` to the `min` from the raster statistic (if present)

vincentsarago: or even one step further 

```python
        if self.nodata_height is not None:
            data[img.array.mask[0]] = (self.nodata_height - self.baseval) / self.interval
        elif img.dataset_statistics:
            # Check image statistics 
            stats = img.dataset_statistics[0]
            data[img.array.mask[0]] = (stats[0] - self.baseval) / self.interval
        else:
            # Get min value from the image (tile)
            data[img.array.mask[0]] = (img.array[0].min() - self.baseval) / self.interval
```

jo-chemla: These defaults do look great indeed in case there's no user-provided nodata_height

jo-chemla:
Weird, will need to check. I've also updated the /cog/viewer endpoint to also support algorithm select and parameters edition, should be good to go!

I'll try to have a look at your feedback tomorrow!

image

vincentsarago:
I think we should decouple the `viewer` part from this PR. 

e.g algorithm might need more than one band, so I don't think they should be in the `1b viz`

vincentsarago:
FYI I've started a PR in rio-viz to support raster-dem viz https://github.com/developmentseed/rio-viz/pull/64 ;-)

jo-chemla:
Wow that was fast, 3d maplibre support looking nice! I can revert the changes specific to the viewer in that titiler PR, should I port the algorithms selection to another PR in Rio-viz instead? Still a wip, it can indeed be useful to other bands as well, will need to strengthen that wip.

vincentsarago:
> should I port the algorithms selection to another PR in Rio-viz instead? Still a wip, it can indeed be useful to other bands as well, will need to strengthen that wip.

@jo-chemla I've personally don't have a lot of time to work on this (I've already spend too much time 🙈) so I would say you can but don't expect too much 😅

jo-chemla:
Alright, thanks again for getting back! I've reverted commits dedicated to `/cog/viewer`. 

Trying to gather what's missing before merging: 
 - testing with `https://data.geo.admin.ch/ch.swisstopo.swissalti3d/swissalti3d_2019_2573-1085/swissalti3d_2019_2573-1085_0.5_2056_5728.tif ` and understand the issue
 - update tests in https://github.com/developmentseed/titiler/blob/57b44d6f2fc32153705550643eb61913f524935f/src/titiler/core/tests/test_algorithms.py#L69-L107 to account for new options
 - pass the tests

I think I can PR the /cog/viewer in a new PR almost as-is, and 
 - activate that `algorithm-section` in only update-1b first, but filter only algos that apply to single-band data
 - then I can also push the algorithms-section to multi-bands, but there might be some duplicated code
 - after your comment on terrain in rio-viz, I thought rio-viz was the module/template behind the `cog_viewer.html` endpoint, but after inspection they seem unrelated - or at least code and changes made in one have to be ported to the other, right? In that case I might focus on the titiler `/cog/viewer` only, and in the future we might port the changes to rio-viz, is that correct?

vincentsarago:
🙏 @jo-chemla 

> after your comment on terrain in rio-viz, I thought rio-viz was the module/template behind the cog_viewer.html endpoint, but after inspection they seem unrelated - or at least code and changes made in one have to be ported to the other, right? In that case I might focus on the titiler /cog/viewer only, and in the future we might port the changes to rio-viz, is that correct?

Yeah they are different templates (because at the beginning they were quite different but with time they evolved to be a bit too close now 😓 

let's not worry about rio-viz for now, you can focus on titiler cog viewer 🙏

jo-chemla:
Just had a quick look, 
 - Seems that the error with the swisstopo dem was just a trailing space, it works for me with the /cog/preview & tiles endpoints, with or without algorithm specified as url-param
 - current tests do pass with `python -m pytest src/titiler/core --cov=titiler.core --cov-report=xml --cov-append --cov-report=term-missing`, with a coverage that can be improved here 
```
Name                                                   Stmts   Miss Branch BrPart  Cover   Missing
src\titiler\core\titiler\core\algorithm\dem.py           123      3     10      5    94%   50->61, 102->113, 182, 236, 239
```

The missing tests are:
 - on pre-existing missing-branches `if self.buffer:`
 - and newly, non-covered branches are the 2 added tests `if self.nodata_height is not None:`. I've added a test where `nodata_height` url-param is passed, with a mask, and check that the values of the output in the mask area equate the provided height. 
 
The thing is I'm not sure which test is best among the 2 solutions below. Either edit `test_terrain_algo` with a client.get response where I add the url-param, or rather edit the `test_terrarium` methods passing the option to the python method. Just committed, and test do pass, you tell me. 
 
```py
# SOLUTION 1
def test_terrain_algo():
    #...
    #def main(algorithm=Depends(default_algorithms.dependency)):
    # client = TestClient(app)
    #   ...

    # SOLUTION 1
    # test nodata_height 
    # MAPBOX Terrain RGB
    response = client.get("/", params={"algorithm": "terrainrgb", "algorithm_params":json.dumps({"nodata_height":10.0})})
    assert response.status_code == 200
    with MemoryFile(response.content) as mem:
        with mem.open() as dst:
            data = dst.read().astype(numpy.float64)
    elevation = -10000 + (((data[0] * 256 * 256) + (data[1] * 256) + data[2]) * 0.1)
    numpy.testing.assert_array_equal(elevation, arr[0])

    # TILEZEN Terrarium
    response = client.get("/", params={"algorithm": "terrarium", "algorithm_params":json.dumps({"nodata_height":10.0})})
    assert response.status_code == 200
    with MemoryFile(response.content) as mem:
        with mem.open() as dst:
            data = dst.read().astype(numpy.float64)
    elevation = (data[0] * 256 + data[1] + data[2] / 256) - 32768
    numpy.testing.assert_array_equal(elevation, arr[0])

# SOLUTION 2
def test_terrarium():
    # ...

    # nodata_height
    nodata_height = 10.0
    algo = default_algorithms.get("terrarium")(nodata_height=nodata_height)
    arr = numpy.ma.MaskedArray(
        numpy.random.randint(0, 5000, (1, 256, 256), dtype="uint16"),
        mask=numpy.zeros((1, 256, 256), dtype="bool"),
    )
    arr.mask[0, 0:100, 0:100] = True
    img = ImageData(arr)
    out = algo(img)
    masked = out.array[:, arr.mask[0,:,:]]
    masked_height = (masked[0] * 256 + masked[1] + masked[2] / 256) - 32768
    print(masked_height.shape)
    numpy.testing.assert_array_equal(masked_height, nodata_height * numpy.ones((100 * 100), dtype="bool"))

```

vincentsarago:
@jo-chemla Solution 2 is fine (which is how the tests are written for now) 🙏 

we don't need to worry to much about the coverage %

jo-chemla:
Thanks, from my point of view we should be ready to go then!
In parallel I've finalized the touchups for the PR dedicated to `/cog/viewer`, which now works for both 1b and 3b afaict, please do tell me if things are not done correctly on the other thread https://github.com/developmentseed/titiler/pull/1119. 

Also, the last item from the original discussion was to generalize algorithms to ingest not only 1b DEMs, but also allow transformation of the input data from terrarium/terrainrgb into altitude/height data (or algorithms chaining) so it could be used on-the-fly by algorithms (hillshade etc). I'm not sure how this could/should work and it can probably be the subject of another PR if deemed useful.

Thanks for all the help as always!

vincentsarago:
@jo-chemla could you run `pre-commit` on your branch? 

```python
python -m pip install pre-commit 
# in titiler root 
pre-commit install 
pre-commit run --all-files
```

> Also, the last item from the original discussion was to generalize algorithms to ingest not only 1b DEMs, but also allow transformation of the input data from terrarium/terrainrgb into altitude/height data (or algorithms chaining) so it could be used on-the-fly by algorithms (hillshade etc). I'm not sure how this could/should work and it can probably be the subject of another PR if deemed useful.

Do you mean go from RGBA (terranium) to Float? (I think you could use band expression for this) 

> algorithms chaining

we could allow algorithm to be a list but then it might be a bit messy when wanting to pass options to each algorithm. 

FYI We've recently worked on OpenEO https://github.com/sentinel-hub/titiler-openeo and the custom [processing](https://openeo.org/documentation/1.0/developers/api/reference.html#section/Processes) language is quite interesting. We might bring this to titiler (if any customer need it 🙈 )

jo-chemla:
Thanks again. Just ran the pre-commit and committed, so should be good to go on this one!

Nice to see the custom Processing language for OpenEO, first time I hear about it 👀. It does look a lot like Node editors for 3D-artist related workflows, eg in case any inspiration could be useful:
 - chaining Blender nodes for [material](https://upload.wikimedia.org/wikipedia/commons/3/3f/Blender_shader_node_editor_showcasing_a_Moldy_Gold_Materiel.png) or geometry, 
 - threejs shading language [TSL](https://github.com/mrdoob/three.js/wiki/Three.js-Shading-Language) NodeMaterial, 
 - or [Alicevision meshroom](https://meshroom-manual.readthedocs.io/en/v19.01.45/node_reference/node_reference.html) processing 
 - [PDAL pipelines](https://pdal.io/en/stable/pipeline.html) is another interesting take for more linear workflows. 

And for algorithms chaining, 🤯 indeed the band expression is the way to prepare the data before algorithm execution (eg the existing aws [open-data terrain tiles](https://registry.opendata.aws/terrain-tiles/) which are stored as terrarium with url `https://s3.amazonaws.com/elevation-tiles-prod/terrarium/${z}/${x}/${y}.png` can be wrapped into a `GDAL_WMS TMS` service passed as url, with `expression=(b1 * 256 + b2 + b3 / 256) - 32768`, and algorithm eg hillshade. 🙈 Full url [here](http://titiler.xyz/cog/preview?url=%3CGDAL_WMS%3E%3CService%20name%3D%27TMS%27%3E%3CServerUrl%3Ehttps%3A%2F%2Fs3.amazonaws.com%2Felevation-tiles-prod%2Fterrarium%2F%24%7Bz%7D%2F%24%7Bx%7D%2F%24%7By%7D.png%3C%2FServerUrl%3E%3C%2FService%3E%3CDataWindow%3E%3CUpperLeftX%3E-20037508.34%3C%2FUpperLeftX%3E%3CUpperLeftY%3E20037508.34%3C%2FUpperLeftY%3E%3CLowerRightX%3E20037508.34%3C%2FLowerRightX%3E%3CLowerRightY%3E-20037508.34%3C%2FLowerRightY%3E%3CTileLevel%3E18%3C%2FTileLevel%3E%3CTileCountX%3E1%3C%2FTileCountX%3E%3CTileCountY%3E1%3C%2FTileCountY%3E%3CYOrigin%3Etop%3C%2FYOrigin%3E%3C%2FDataWindow%3E%3CProjection%3EEPSG%3A3857%3C%2FProjection%3E%3CBlockSizeX%3E256%3C%2FBlockSizeX%3E%3CBlockSizeY%3E256%3C%2FBlockSizeY%3E%3CBandsCount%3E3%3C%2FBandsCount%3E%3CCache%20%2F%3E%3C%2FGDAL_WMS%3E&bidx=1&expression=%28b1%20%2A%20256%20%2B%20b2%20%2B%20b3%20%2F%20256%29%20-%2032768&max_size=1024&algorithm=hillshade)

Makes me want to add expression support before merging #1119

vincentsarago:
Again I don't want to add too much stuff into the `viewer` (mostly because it's written in pure htm/js I did and I'm not a frontend dev 😅) 

If you really want to play/visualize the data I often use the `/map` endpoint directly

jo-chemla:
Alright, just tried adding the `expression` support to `/cog/viewer`, only to 1b section since I could only test it on this. I'll let you decide whether or not you want that addition merged!

vincentsarago (approved):

PR #1134, opened 2025-04-22
refactor xarray dimension selection

### LINKED ISSUE #1099: XArray IO can only select time at daily intervals
#### Problem description

In `xarray.io.get_variable`, if the `datetime` argument is an ISO8601 datetime like `2025-02-25T15:33:38Z` with a _date_ and a _time_, the time part of the datetime is dropped. This causes a problem for assets whose temporal dimension is more frequent than daily, because `get_variable` will always snap to the nearest day.

I think the fix is to simply remove the split by `T` and keep `datetime` intact.

https://github.com/developmentseed/titiler/blob/28d33a620e3f409f3ebf7d1335d91dd6d94719da/src/titiler/xarray/titiler/xarray/io.py#L175-L187

### LINKED ISSUE #1112: [Xarray] Improve dimension selection
Right now the `dimension` selection in titiler-xarray is very limited. We have 2 options:
- `drop_dim={dimension}={value}` (select a value and drop the dimension)

https://github.com/developmentseed/titiler/blob/6d1519e9a289a1f1f8e15c04020920f82cf02c60/src/titiler/xarray/titiler/xarray/io.py#L156-L160
 
- `datetime={date}` 

https://github.com/developmentseed/titiler/blob/6d1519e9a289a1f1f8e15c04020920f82cf02c60/src/titiler/xarray/titiler/xarray/io.py#L175-L187

The `drop_dim` parameter feels kinda weird to me. We are using it to `select a slice` and `drop a dimension`, while the `datetime` is limited to the `time` dimension. 

I'm looking for opinions here but here are my though:

- `drop_dim` should only drop dimension (not select a slice)
- `datetime` should be deprecated
- we should add a `sel={dimension}={value}` option (https://docs.xarray.dev/en/latest/user-guide/indexing.html) 
- we could introduce a `band_dim={dimension}` which will let user select the dimension they want to appear as `bands` in titiler. Resulting in dropping all other dimensions (except X and Y)  

cc @sharkinsspatial @maxrjones @j08lue

vincentsarago:
closes #1112, closes #1099

With the change in this PR a user could now do 

```
# select array for time = 2024-02-10 and with elevation=200
# Xarray: da.sel({"time": "2024-02-10", "elevation": 200}, method="nearest"})
# With Titiler:
{titiler endpoint}/info?url={netcdf.nc}&sel=time=2024-02-10&sel=elevation=200&method=nearest

# Create a 2 band array (for each time dim selection)
# Xarray: da.sel({"time": ["2024-02-10", "2024-02-11"], "elevation": 200}, method="nearest"})
# With Titiler:
{titiler endpoint}/info?url={netcdf.nc}&sel=time=2024-02-10&time=2024-02-11&sel=elevation=200&method=nearest
```

cc @j08lue @maxrjones @sharkinsspatial @hrodmn


src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -173,20 +183,6 @@ def get_variable(
...
-            # TODO: handle time interval
-            time_as_str = datetime.split("T")[0]
-            if da["time"].dtype == "O":
-                da["time"] = da["time"].astype("datetime64[ns]")
-
-            da = da.sel(
-                time=numpy.array(time_as_str, dtype=numpy.datetime64), method="nearest"
-            )
-        else:
-            da = da.isel(time=0)
```
vincentsarago, line 188: ⚠️ we do not select the first time dimension now

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -138,27 +138,37 @@ def _arrange_dims(da: xarray.DataArray) -> xarray.DataArray:
...
+    if sel:
+        sel_idx = []
+        for s in sel:
+            val: Union[str, slice]
+            dim, val = s.split("=")
+            dt = da[dim].dtype
+            if "," in val:
+                val = slice(*[numpy.array(v).astype(dt) for v in val.split(",")])
+            else:
+                val = numpy.array(val).astype(dt)
```
vincentsarago, line 167: we need to `cast` the value to the array datatype (e.g `z="0"` -> `z=0`)

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -138,27 +138,37 @@ def _arrange_dims(da: xarray.DataArray) -> xarray.DataArray:
...
-            dim_to_drop, dim_val = dim.split("=")
-            da = da.sel({dim_to_drop: dim_val}).drop_vars(dim_to_drop)
+    if sel:
+        sel_idx = []
+        for s in sel:
+            val: Union[str, slice]
+            dim, val = s.split("=")
+            dt = da[dim].dtype
+            if "," in val:
+                val = slice(*[numpy.array(v).astype(dt) for v in val.split(",")])
```
vincentsarago, line 165: I took a chance here, would a value have `,`? 🤷 

basically this is to support things like `time=date1,date2` -> da.sel(time=slice(date1, date2))`

j08lue: You could branch on the dimension dtype being string and in that case not do the slice. 🤷 

Otherwise, I cannot think of a case where a value would contain comma. So perhaps just leave as-is until someone hits this issue.

src/titiler/xarray/titiler/xarray/io.py
j08lue: Do we? in the `slice` case, yes. But otherwise, I think Xarray / Pandas does this casting for you, no?

j08lue: Does this work well for datetime dtype?

For datetime dimension dtype, you could also not cast the indices (do the comma-split to slices, though) and leave the casting to Pandas, preserving the nice magic is has: https://github.com/developmentseed/titiler/issues/1112#issuecomment-2761260478

src/titiler/xarray/titiler/xarray/io.py
vincentsarago: 😓 I guess we need to come up with something else.

j08lue: Nah, I think it's fair to expect that people do not use the most [obscure](https://xkcd.com/1179/) ways of specifying datetime.

So I'd stick to the splitting on comma - just raise good errors when it fails.

vincentsarago: we could also use `slice` is user pass twice the same dim `time=date1,time=date2` -> `time=slice(date1, date2)` 🤷‍♂️

src/titiler/xarray/titiler/xarray/io.py
vincentsarago: 🤦 we need to do like xpublish https://github.com/xpublish-community/xpublish-wms/blob/15aaa48ead7468f80d4794456b674a28bbc7dbe0/xpublish_wms/wms/get_map.py#L251-L269

src/titiler/xarray/titiler/xarray/io.py
j08lue: Wait, why do we do slices at all? Apart from maybe the spatial dimensions, you always need to reduce the dimensions to 1, i.e. select a single index, no?

Passing the same dim twice would, to me, mean `sel(time=[date1, date2])` - but same thing here re reducing dimensionality.

Or are you planning to treat multiple 2D fields as "bands"?

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -135,30 +136,53 @@ def _arrange_dims(da: xarray.DataArray) -> xarray.DataArray:
...
+            val: Union[str, slice]
+            dim, val = s.split("=")
+            val = _cast_to_type(val, da[dim].dtype)
+
+            if dim in _idx:
+                _idx[dim].append(val)
+            else:
+                _idx[dim] = [val]
+
+        sel_idx = {k: v[0] if len(v) < 2 else slice(*v) for k, v in _idx.items()}
```
vincentsarago, line 184: if a `dim` is provided multiple times we transform values to `slice`

j08lue: Sounds like a good API design choice, but should we actually be supporting slicing? https://github.com/developmentseed/titiler/pull/1134#discussion_r2055602147

src/titiler/xarray/tests/test_factory.py
```diff
@@ -165,21 +165,21 @@ def test_tiles(filename, app):
 
     resp = app.get(
         "/md/tiles/WebMercatorQuad/0/0/0",
-        params={"url": filename, "variable": "dataset", "rescale": "0,500"},
+        params={"url": filename, "variable": "dataset", "rescale": "0,500", "bidx": 1},
```
hrodmn, line 168: why do you need to add `bidx=1` now? Is it because the reader will return a 3D array now with the new `sel` interface (instead of 2D with `drop_dim`)?

src/titiler/xarray/titiler/xarray/dependencies.py
```diff
@@ -71,17 +74,22 @@ class CompatXarrayParams(XarrayIOParams):
...
     """
 
-    variable: Annotated[Optional[str], Query(description="Xarray Variable name")] = None
+    variable: Annotated[Optional[str], Query(description="Xarray Variable name.")] = (
+        None
+    )
 
-    drop_dim: Annotated[
-        Optional[List[DropDimStr]],
+    sel: Annotated[
```
hrodmn, line 81: this is so cool!

hrodmn (commented):
I really like bringing the `sel` interface to the titiler endpoints - it's easy to imagine how you could build a UI that combines the `/dataset/info` response with some user input to generate tile URLs.

src/titiler/xarray/tests/test_factory.py
j08lue: Same question as I asked, basically: https://github.com/developmentseed/titiler/pull/1134#discussion_r2055602147

src/titiler/xarray/titiler/xarray/io.py
vincentsarago: > Or are you planning to treat multiple 2D fields as "bands"?

Yes we do support 3D data now, meaning you could have multiple Z (time or dim) represented as bands

src/titiler/xarray/tests/test_factory.py
vincentsarago: > why do you need to add bidx=1 now? Is it because the reader will return a 3D array now with the new sel interface (instead of 2D with drop_dim)?

Exactly, also before we were always selecting the first time value. Now we can have 3D arrays

src/titiler/xarray/titiler/xarray/io.py
j08lue: Hmm, coming back to this, I think I would as an API user really expect multiple `sel` URL parameters to select those exact levels, instead of constructing a slice.

I think you should drop the `slice` support altogether. Here is why:

1. Unpredictable "bands" - If you want to get multiple bands (e.g. for band math), you want to know exactly what indexes they correspond to. Slicing can give a lot of surprises - fewer or more indexes included in the slice than you expected, maybe there was a level missing that you thought was there, etc.
2. Slicing without reducing is not really useful - I cannot think of a use case from my NetCDF / data cube work where I would do slices but then not reduce the data over that dimension (e.g. averaging a time slice).
3. TiTiler ultimately needs 2D - the user always needs to reduce the data to a 2D field, if not through singular indexing, than through "band" math (aka expressions), I guess. There are few workflows where slicing would be part of a predictable process of getting to a 2D field.

Do you have strong reasons for supporting slicing, @vincentsarago?

vincentsarago: > Unpredictable "bands" - If you want to get multiple bands (e.g. for band math), you want to know exactly what indexes they correspond to. Slicing can give a lot of surprises - fewer or more indexes included in the slice than you expected, maybe there was a level missing that you thought was there, etc

FYI we do not support band math, only band selection. 

> you want to know exactly what indexes they correspond to

If you hit the `/info` endpoint first you would know the band indexes 

> Slicing without reducing is not really useful - I cannot think of a use case from my NetCDF / data cube work where I would do slices but then not reduce the data over that dimension (e.g. averaging a time slice).

My idea was that sometime you don't know exactly which value you want and `slicing` might be better than `nearest` selection. We could also think about user doing `slice` selection and then reducing using algorithm (mean/sum....) 

> TiTiler ultimately needs 2D - the user always needs to reduce the data to a 2D field, if not through singular indexing, than through "band" math (aka expressions), I guess. There are few workflows where slicing would be part of a predictable process of getting to a 2D field.

Not really you want 3D data where you can do band combination or apply algorithm

vincentsarago: Summary of our discussion on meet. 

- We should drop `slice` 
- user might want to select multiple values e.g `date1` and `date3` without selecting `date2` (`time=date1&time=date3`)

j08lue (approved):
Nice, this is a very powerful feature - and with the new naming and behavior it will be very familiar to Xarray users.

Should we rename the `method` parameter to indicate that it is specific to the `sel` operation, e.g. `sel_method`? 🤔 Just a thought.

maxrjones:
Amazing improvement! What if someone wants to use different methods along different dimensions? E.g., I could imagine wanting nearest in space but exact in time. IMO it would make more sense to include method in the param for sel and split on ‘method=‘

vincentsarago:
@maxrjones 

> Amazing improvement! What if someone wants to use different methods along different dimensions? E.g., I could imagine wanting nearest in space but exact in time. IMO it would make more sense to include method in the param for sel and split on ‘method=‘

Interesting, for now we are not chaining the `sel` operations but using `dict` to pass everything to one operation so the method would be the same for all the `dim` 

I'm a bit worry about over complexity of `method per dim` 🤔

src/titiler/xarray/tests/test_factory.py
```diff
@@ -135,22 +135,39 @@ def test_info_da_options(app):
     """Test /info endpoints with Dataarray options."""
     resp = app.get(
         "/md/info",
-        params={"url": dataset_4d_nc, "variable": "dataset", "drop_dims": "z=0"},
+        params={"url": dataset_4d_nc, "variable": "dataset", "sel": "time=2023-01-01"},
     )
     assert resp.status_code == 200
     assert resp.headers["content-type"] == "application/json"
+    body = resp.json()
+    print(body)
```
abarciauskas-bgse, line 143: nit: probably don't intend to keep the print statement here?

vincentsarago: 😓 🙏

maxrjones:
> @maxrjones
> 
> > Amazing improvement! What if someone wants to use different methods along different dimensions? E.g., I could imagine wanting nearest in space but exact in time. IMO it would make more sense to include method in the param for sel and split on ‘method=‘
> 
> Interesting, for now we are not chaining the `sel` operations but using `dict` to pass everything to one operation so the method would be the same for all the `dim`
> 
> I'm a bit worry about over complexity of `method per dim` 🤔

that makes sense. I didn't test this your PR out thoroughly but generally really like the approach of using `sel`.

vincentsarago:
@maxrjones I had more though about `chaining` sel and I don't think it will be possible . 

I think I'm going to merge this PR as is right now but we could re-visit the logic in `titiler-multidim`.  `titiler.xarray` will provide a simple default behaviour and sub-libraries can explore more complex ways to access the data which we then can bring back here after user feedback.

Issue #1126, opened 2025-04-09
add `/map` prefix to tiles endpoint


vincentsarago:
right now we have tiles endpoints in form of `/tiles/{tileMatrixSetId}/{z}/{x}/{y}` which is compliant with OGC Tileset specification.

we could also comply with `dataset-tileset` and `tilesets-map` conformance  by adding `/map` prefix to the /tiles endpoints 

```
# now                               
../tiles                                                   
../tiles/{tileMatrixSetId}                                     
../tiles/{tileMatrixSetId}/{z}/{x}/{y}[@{scale}x][.{format}]
/tileMatrixSets
/tileMatrixSets/{tileMatrixSetId}

ConformsTo:
- http://www.opengis.net/spec/ogcapi-tiles-1/1.0/conf/core
- http://www.opengis.net/spec/ogcapi-tiles-1/1.0/conf/tileset
- http://www.opengis.net/spec/ogcapi-tiles-1/1.0/conf/tilesets-list
- http://www.opengis.net/spec/ogcapi-tiles-1/1.0/conf/oas30

# after                                  
../map/tiles                                                   
../map/tiles/{tileMatrixSetId}                                     
../map/tiles/{tileMatrixSetId}/{z}/{x}/{y}[@{scale}x][.{format}] 
/tileMatrixSets
/tileMatrixSets/{tileMatrixSetId}

ConformsTo: 
- http://www.opengis.net/spec/ogcapi-tiles-1/1.0/conf/core
- http://www.opengis.net/spec/ogcapi-tiles-1/1.0/conf/tileset
- http://www.opengis.net/spec/ogcapi-tiles-1/1.0/conf/tilesets-list
- http://www.opengis.net/spec/ogcapi-tiles-1/1.0/conf/dataset-tilesets
- http://www.opengis.net/spec/ogcapi-tiles-1/1.0/conf/tilesets-map
- http://www.opengis.net/spec/ogcapi-tiles-1/1.0/conf/oas30
``` 

ref: https://docs.ogc.org/is/20-057/20-057.html

j08lue:
Does TiTiler-PgSTAC have a concept of "datasets" ("a.k.a., a root resource, which may contain multiple geospatial data resources")? I guess it would in our case just be a transparent wrapper to the existing resources (collections, tilesets)?

In other words, would the addition of `/map` provide anything new and useful, or just an alias, for compliance?

j08lue:
[...]

Issue #753, opened 2024-01-16
OGC API - Maps 


vincentsarago:
> OGC API - MAPS
> 
> The OGC API - Maps draft specification describes an API that can serve spatially referenced and dynamically rendered electronic maps. The specification describes the discovery and query operations of an API that provides access to electronic maps in a manner independent of the underlying data store. The query operations allow dynamically rendered maps to be retrieved from the underlying data store based upon simple selection criteria, defined by the client.

https://ogcapi.ogc.org/maps/

https://docs.ogc.org/DRAFTS/20-058.html

Same as OGC Tiles and Coverage, we won't support the spec to the letter but there might be endpoints we can create in an extension.

I need to look a bit more but to me the specs looks like the IIIF spec which we played with in https://github.com/developmentseed/titiler-image

jerstlouis:
@vincentsarago The standard is very modular and the requirements for the ["Core" class](https://docs.ogc.org/DRAFTS/20-058.html#rc_table_core) are very few (`.../map` returns a map with a `Content-Bbox:` header, and `Content-Crs:` if not in OGC CRS84), so it should be easy to conform to "Core".

Then you could gradually implement the additional requirement classes you care about (e.g, adding support for query parameters like [`bbox=`](https://docs.ogc.org/DRAFTS/20-058.html#rc_table_spatial-subsetting) and [`width=`/`height=`](https://docs.ogc.org/DRAFTS/20-058.html#rc_table_scaling)).

kalxas:
There is now this pygeoapi provider using titiler under the hood:
https://github.com/EOEPCA/eoapi-maps-plugin

PR #1136, opened 2025-04-23
rename /map to /map.html

vincentsarago:
ref https://github.com/developmentseed/titiler/issues/753 https://github.com/developmentseed/titiler/issues/1126#issuecomment-2789156024

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Misc
+
+* rename `/map` endpoint to `/map.html` **breaking change**
+
 ### titiler.application
 
 * fix Landing page links when app is behind proxy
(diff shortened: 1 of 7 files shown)


hrodmn (approved):

j08lue (changes requested):
Since we are renaming this anyways (breaking change), would it perhaps be better to choose a less similar name to `/map` and perhaps even more descriptive?

Something like `/preview.html` or `/viewer.html` or so?

vincentsarago:
@j08lue why not. 

we already have `/preview` so `preview.html` will be confusing. There is already a `/viewer` which is provided by the VIewerExtension (return a much more complexe HTML viewer).

j08lue:
> we already have [`/preview` and `/viewer`]

Right, I forgot about that. Well, I do not have any other ideas, so ¯\_(ツ)_/¯

j08lue (approved):

PR #1148, opened 2025-05-02
add `/validate` endpoint for md dataset

vincentsarago:
Q:
- should we validate the entire dataset 
- should we validate all the `variables` 
- if a dataset if valide, does all the variables are? 

My knowledge of Xarray dataset structure is limited so would love some help on this 🙏

### DIFF
--- a/src/titiler/xarray/titiler/xarray/extensions.py
+++ b/src/titiler/xarray/titiler/xarray/extensions.py
@@ -15,7 +15,7 @@ from titiler.core.factory import FactoryExtension
 from titiler.core.resources.enums import MediaType
 from titiler.xarray.dependencies import XarrayIOParams
 from titiler.xarray.factory import TilerFactory
-from titiler.xarray.io import open_zarr
+from titiler.xarray.io import X_DIM_NAMES, Y_DIM_NAMES, open_zarr
 
 
 @define
@@ -139,12 +139,8 @@ class ValidateExtension(FactoryExtension):
 
         if "y" not in da.dims:
             try:
-                latitude_var_name = next(
-                    name
-                    for name in ["lat", "latitude", "LAT", "LATITUDE", "Lat"]
-                    if name in da.dims
-                )
-                da = da.rename({latitude_var_name: "y"})
+                y_dim = next(name for name in Y_DIM_NAMES if name in da.dims)
+                da = da.rename({y_dim: "y"})
 
             except StopIteration:
                 errors.append(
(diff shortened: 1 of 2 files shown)


maxrjones:
> Q:
> 
> * should we validate the entire dataset
> * should we validate all the `variables`
> 
> My knowledge of Xarray dataset structure is limited so would love some help on this 🙏

Could specific data variable be included as a query parameter and all data variables are validated if not a specific one is not queried? E.g., if someone knows they will want to tile `precipitation` they would likely not want to go through the effort to validate all the DataArrays in the Dataset.

> * if a dataset if valide, does all the variables are?

No, as one common example the CRS information when using rioxarray is often stored in a `crs` data variable which is not valid for titiler even though the dimensions and coordinates of the dataset are valid:
image

vincentsarago:
🙏 @maxrjones 

Right now the I have added the `variable=` option which will validate a specific variable, if not present it will try to validate the whole dataset. 

If I understand well, when variable is not set, you'll iterate through all the variables and return something like 
```
{
    "var1": {
        "compatible_with_titiler": true,
        "errors": []
    },
    "var2": {
        "compatible_with_titiler": false,
        "errors": [
            "Dataset does not have compatible spatial coordinates"
        ]
    },
    ...
}
```

maxrjones:
This approach seems good for short-term use, though eventually it would be nice to use a schema validation approach (e.g., https://github.com/xarray-contrib/xarray-schema) rather than eagerly applying the operations to save time and memory.

My main concern is about the validation and io code becoming out of sync. Could one or more utility functions be defined for some of the code in https://github.com/developmentseed/titiler/blob/main/src/titiler/xarray/titiler/xarray/io.py#L96-L202 for use in both io and validate to mitigate that chance?

It also looks like the metadata and variable extensions are becoming redundant. Is the separate Variable extension still needed?

vincentsarago:
updated with a `variables=...` options (allowing multiple variables), if not it will run the validation for all variables 

```
http://127.0.0.1:8080/validate?url=https://nasa-power.s3.us-west-2.amazonaws.com/syn1deg/temporal/power_syn1deg_monthly_temporal_lst.zarr&variables=AIRMASS_00&variables=AIRMASS_01

{
  "AIRMASS_00": {
    "compatible_with_titiler": true,
    "errors": [],
    "warnings": []
  },
  "AIRMASS_01": {
    "compatible_with_titiler": true,
    "errors": [],
    "warnings": []
  }
}
```

PR #1150, opened 2025-05-06
fix: use minZoom instead of minNativeZoom

hrodmn:
By setting `minNativeZoom` we were constructing low zoom tiles from many higher-zoom tile requests! This is not the intended effect of the minzoom setting. We can still overzoom by using maxNativeZoom.


vincentsarago (approved):
🚀

PR #1155, opened 2025-05-13
[titiler.xarray] update type informations for reader and path_dependencies

vincentsarago:
ref https://github.com/developmentseed/titiler/discussions/1154#discussioncomment-13131173

### DIFF
--- a/src/titiler/xarray/titiler/xarray/factory.py
+++ b/src/titiler/xarray/titiler/xarray/factory.py
@@ -1,12 +1,13 @@
 """TiTiler.xarray factory."""
 
-from typing import Callable, Optional, Type, Union
+from typing import Any, Callable, Optional, Type, Union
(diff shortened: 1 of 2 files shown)


src/titiler/xarray/titiler/xarray/factory.py
```diff
@@ -32,9 +33,9 @@
 class TilerFactory(BaseTilerFactory):
     """Xarray Tiler Factory."""
 
-    reader: Type[Reader] = Reader
+    reader: Type[XarrayReader] = Reader
 
-    path_dependency: Callable[..., str] = DatasetPathParams
+    path_dependency: Callable[..., Any] = DatasetPathParams
```
vincentsarago, line 38: we could have `Union[str, xarray.DataArray]`

PR #1164, opened 2025-05-29
fix xarray_open_dataset for cloud hosted files

vincentsarago:


vincentsarago:
@maxrjones are we good to merge here?

maxrjones (approved):
thanks!

PR #1165, opened 2025-06-03
fix: rename variance algorithm

Skaidus:
Renamed band name for the variance algorithm ~~probably std copypaste~~


vincentsarago (approved):

vincentsarago:
🙏 thanks @Skaidus

PR #1171, opened 2025-06-12
add OpenTelemetry instrumentation to titiler.core

hrodmn:
From #1167 

This adds some convenience functions for adding opentelemetry traces to titiler applications in titiler.core.telemetry. These features will only be activated if the opentelemetry dependencies are installed.

When enabled in a titiler application, the traces can be browsed in a UI like Jaeger:
image

**Factory class instrumentation**: the `BaseTilerFactory` has a new method `add_telemetry()` which will add a trace to each registered endpoint in the factory class.

**Endpoint instrumentation**: endpoint methods in all of the tiler factory classes in `titiler.core`, `titiler.mosaic`, and `titiler.xarray` have been decorated with some `operation_tracer` context managers to track the performance of specific operations within an endpoint function.

To keep things clean I opted to skip adding additional attributes to the traces, but we might want to revisit that in the future!

### TODO
- [x] add telemetry.py
- [x] decorate factory classes
- [x] add tests
- [x] update docs


src/titiler/application/pyproject.toml
```diff
@@ -50,6 +50,10 @@ test = [
 server = [
     "uvicorn[standard]>=0.12.0",
 ]
+telemetry = [
```
vincentsarago, line 53: so `titiler.application` is a **demo** application so IMO it's fine if it comes with `battery included` meaning that I'll be fine to have `telemetry` packages shipped with it by default. 

if users do not want the telemetry packages they can easily build their own application with `titiler.core` without needing `titiler.application`

src/titiler/core/titiler/core/factory.py
```diff
@@ -851,20 +873,47 @@ def tile(
...
-                        **dataset_params.as_dict(),
-                    )
-                    dst_colormap = getattr(src_dst, "colormap", None)
+            with trace_operation(
+                "open_dataset",
+                {
+                    "dataset.path": src_path,
+                    "dataset.tms": tileMatrixSetId,
+                },
+            ) as span:
```
vincentsarago, line 882: `io_span`?

src/titiler/core/titiler/core/telemetry.py
```diff
@@ -0,0 +1,206 @@
...
+import functools
+import inspect
+import sys
+from contextlib import contextmanager
+from typing import Any, Callable, Dict, Iterator, Optional, TypeVar, overload
+
+if sys.version_info >= (3, 10):
+    from typing import ParamSpec
+else:
+    from typing_extensions import ParamSpec
```
vincentsarago, line 12: I think typing extensions already has the `if sys.version_info` check, so while we make python>=3.10 we can just use `from typing_extensions import ParamSpec`

src/titiler/core/titiler/core/telemetry.py
```diff
@@ -0,0 +1,206 @@
...
+        return decorator
+    else:
+        return decorator(_func)
+
+
+factory_trace.decorator_enabled = bool(tracer)  # type: ignore [attr-defined]
+
+
+@contextmanager
+def trace_operation(
```
vincentsarago, line 186:
```suggestion
def trace_operation(
```
```suggestion
def tracer(
```

vincentsarago:
Thanks @hrodmn this looks super promising 🙏 

I've added couple comments (mostly on naming).

My main concern is about the number of lines the tracing will add to the factory, mostly due to the `set_attribute` blocks which we should try to avoid if the information can be derived from other place (e.g by looking at the query parameter, like  for dataset path and tile matrix set)

If we move forward, we would need to add trace to all other factory endpoints 😬

hrodmn:
We could decide to completely skip setting attributes for  and just add the `operation_tracer` context manager in places where we want more detailed profiling or performance evaluation.

We can still get a trace timeline like this without exporting all of the extra attributes:
image

src/titiler/core/titiler/core/factory.py
```diff
@@ -851,20 +873,47 @@ def tile(
...
+                            image = src_dst.tile(
+                                x,
+                                y,
+                                z,
+                                tilesize=scale * 256,
+                                **tile_params.as_dict(),
+                                **layer_params.as_dict(),
+                                **dataset_params.as_dict(),
+                            )
+                            read_span.set_attributes(
```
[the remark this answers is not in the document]
hrodmn: For now we will skip setting attributes in the spans

hrodmn:
@vincentsarago thanks for your feedback on these changes! I have implemented your naming suggestions and trimmed out the verbose attribute tracing lines.

I did some research on the expected performance hit for including these traces and I don't think it will be significant. Using the `BatchSpanProcessor` reduces overhead for an application because it collects traces in batches and sends them to the collector endpoint in a separate thread. It is also conventional for an application to only collect traces for a sample (e.g. 2%) of requests, which would limit the impact of enabling telemetry if there is any measureable performance reduction.

hrodmn:
@vincentsarago I think this is ready for another look. The approach I have taken is to automatically add OTEL spans to factory methods which will be helpful for easily separating titiler operations from FastAPI operations in traces if the FastAPI Instrumentor is enabled. There are not any within-method spans anymore, but I added a few logging statements to the factory methods. I am not sure how useful those `'opening dataset with reader: <Reader>'` messages will be so we could take them out, but I thought adding some more visibility into the specific reader class that is getting used could be helpful in some contexts.

I did some more digging on OTEL's performance and it will be interesting to see what the impact is in AWS Lambda deployments - reviews are mixed. It is possible that enabling OTEL will have a noticeable impact on cold start times but I have not yet tested it yet.

src/titiler/core/titiler/core/middleware.py
```diff
@@ -107,29 +109,41 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send):
...
+            "http.request.header.accept-encoding": request.headers.get(
+                "accept-encoding"
+            ),
+            "http.request.header.origin": request.headers.get("origin"),
+            "net.host.name": request.url.hostname,
+            "net.host.port": request.url.port,
+            "titiler.query_params": dict(request.query_params),
         }
 
+        telemetry.add_span_attributes(telemetry.flatten_dict(data))
```
hrodmn, line 135: we need to add the span attributes `before` awaiting the app, otherwise these attributes will not get attached to any span because it will already be closed.

src/titiler/core/titiler/core/factory.py
```diff
@@ -225,6 +233,20 @@ def add_route_dependencies(
...
                 route.dependencies.extend(dependencies)  # type: ignore
 
+    def add_telemetry(self):
+        """
+        Applies the factory_trace decorator to all registered API routes.
+
+        This method iterates through the router's routes and wraps the endpoint
+        of each APIRoute to ensure consistent OpenTelemetry tracing.
+        """
+        if not factory_trace.decorator_enabled:
```
vincentsarago, line 243: What about having a setting directly in the Factory and then use it in L178 such as
```python
if self.enable_telemetry:
    self.add_telemetry()
```

src/titiler/core/titiler/core/factory.py
```diff
@@ -852,6 +874,7 @@ def tile(
             """Create map tile from a dataset."""
             tms = self.supported_tms.get(tileMatrixSetId)
             with rasterio.Env(**env):
+                logger.info(f"opening data with reader: {self.reader}")
```
vincentsarago, line 877: > I am not sure how useful those 'opening dataset with reader: <Reader>' messages will be so we could take them out, but I thought adding some more visibility into the specific reader class that is getting used could be helpful in some contexts.

Yeah, we can keep them (maybe remove `with reader: {self.reader}`), I'll be fine to add more logger statement for the `reading` part and rendering, it will be a lot of logging step to add but 🤷

hrodmn: I like having the `reader` class logged because it can help a user understand how the tiles are actually being generated. You can get this info from a traceback but if you don't hit an error than it is not always clear!

I could add `reader`/`backend` attributes to telemetry traces instead if you would rather leave the reader class out of the actual log messages.

src/titiler/core/titiler/core/factory.py
hrodmn: We could do that to provide additional control to the user rather than just turning it on if the telemetry dependencies are installed. I'll add it in.

src/titiler/core/titiler/core/middleware.py
```diff
@@ -107,29 +109,41 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send):
...
-            await self.app(scope, receive, send)
-        except Exception as e:
-            exception = e
-
         request = Request(scope, receive=receive)
-
         data = {
-            "method": request.method,
-            "referer": next(
+            "http.method": request.method,
```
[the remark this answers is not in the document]
hrodmn: Yeah, structured logging looks great. I filed a follow-up: https://github.com/developmentseed/titiler/issues/1190

PR #1172, opened 2025-06-13
allow timezone aware time selection

### LINKED ISSUE #1170: _cast_to_type should support datetime types and not infer timezone
We need to support `sel=time=<datetime>` in titiler.xarray so that we can select within the time dimension for a given xarray dataset (NetCDFs with multiple timestamps or Zarr, etc).

For this we need `_cast_to_type` to support casting the string parameter to a datetime type. With a naive implementation (pandas.to_datetime) when the data type included "datetime", e.g.:

```python
if "datetime" in str(dtype):
    return pandas.to_datetime(value)
```

and then using the sel parameter, we will get an error when the value contains timezone information: the value is timezone aware (i.e. of type `np.datetime[ns, UTC]`) cannot be compared with the timezone-agnostic xarray np.datetime64 series (type `np.datetime[ns]`.

From what I can tell so far, xarray only supports timedelta64 and timezone-agnostic datetime64 data types. So I think we can use explicit casting `np.datetime64(value)`.

```python
elif dtype == numpy.timedelta64:
    return numpy.timedelta64(value)
```

Curious if there are other things we should consider in this implementation.

cc @maxrjones @hrodmn

vincentsarago:
closes #1170 

### This PR does:

Convert any `time` dimension selection parameter to `numpy.datetime64` which will return time in UTC using `value = pandas.Timestamp(value).to_datetime64()` 

We assume all input dataset will have their `<time>` dimension as UTC (because xarray only support numpy.datetime64) 

ref https://github.com/developmentseed/titiler/issues/1170#issuecomment-2968560888

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### titiler.xarray
+
+* add support for timezone aware time selection (e.g `sel="time=2023-01-01T00:00:00+03:00"`)
+
 ## 0.22.2 (2025-06-02)
 
 ### titiler.application
--- a/src/titiler/xarray/tests/test_io_tools.py
+++ b/src/titiler/xarray/tests/test_io_tools.py
@@ -169,6 +169,46 @@ def test_get_variable():
         da = get_variable(ds, "dataset")
 
 
+def test_get_variable_datetime_tz():
+    """test io.get_variable with datetime and timezones."""
+    arr = numpy.arange(0, 33 * 35 * 2).reshape(2, 33, 35)
+    data = xarray.DataArray(
+        arr,
+        dims=("time", "y", "x"),
+        coords={
+            "x": numpy.arange(-170, 180, 10),
+            "y": numpy.arange(-80, 85, 5),
+            "time": [
+                datetime(2022, 1, 1),
+                datetime(2023, 1, 1),
+            ],
+        },
+    )
+    data.attrs.update({"valid_min": arr.min(), "valid_max": arr.max()})
+    assert not data.rio.crs
+    assert data.dims == ("time", "y", "x")
+    ds = data.to_dataset(name="dataset")
+
+    da = get_variable(ds, "dataset", sel=["time=2023-01-01T00:00:00"], method="nearest")
+    assert da.rio.crs
+    assert da.dims == ("y", "x")
+    assert da["time"] == numpy.datetime64("2023-01-01")
+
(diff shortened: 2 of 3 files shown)


abarciauskas-bgse:
@vincentsarago Thank you for opening this change! A few questions:

1. I don't understand the comment `I think there is a major difference with what was implemented in titiler-cmr.` - are you referring to https://github.com/developmentseed/titiler-cmr/pull/52? That PR just integrates upgrades to use the `sel` parameter to pass a single datetime, which is all we need for now (e.g. no slice/range, we can use the datetime parameter for that)
2. What is the reason we can't use `da[dim].dtype.type(val)` in place of _cast_to_type like I've done https://github.com/developmentseed/titiler/pull/1168? I'm fine with this implementation if it works and it is your preference, just want to understand why we need the extra logic in _cast_to_type. I see there is a test failing for `1st of January 2023` but curious if titiler.xarray should have to support that format of date string.

vincentsarago:
> I don't understand the comment I think there is a major difference with what was implemented in titiler-cmr. - are you referring to https://github.com/developmentseed/titiler-cmr/pull/52? That PR just integrates upgrades to use the sel parameter to pass a single datetime, which is all we need for now (e.g. no slice/range, we can use the datetime parameter for that)

Looking at titiler cmr I thought I've seen some parameter like `datetime={datetime_start}/{datetime_end}` but I may be mistaking. We're all good then 👍 

> What is the reason we can't use da[dim].dtype.type(val) in place of _cast_to_type like I've done https://github.com/developmentseed/titiler/pull/1168? I'm fine with this implementation if it works and it is your preference, just want to understand why we need the extra logic in _cast_to_type. I see there is a test failing for 1st of January 2023 but curious if titiler.xarray should have to support that format of date string.

In theory `da[dim].dtype.type(val)` could work but is limited in term of datetime format support (in comparison to `pandas.Timestamp`. 

```python
# With Numpy
numpy.datetime64("2023/01/01")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[7], line 1
----> 1 numpy.datetime64("2023/01/01")

ValueError: Error parsing datetime string "2023/01/01" at position 4

# with Pandas
pandas.Timestamp("2023/01/01").to_datetime64()
np.datetime64('2023-01-01T00:00:00')
```

If we think we don't need to support to many date format I'll be fine switching to `da[dim].dtype.type(val)`

PR #1239, opened 2025-10-15
fix: OpenAPI examples

guillemc23:
This should fix the error when generating clients from Titiler's OpenAPI spec. https://github.com/developmentseed/titiler/issues/1240

Tested generating clients for TypeScript and Python. Python still raises an error but it is due to some incompatibility with Pydantic, but the TypeScript client is succesfully generated.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,12 +5,12 @@
 ### titiler.core
 
 * update `TileJSON` spec from 2.2.0 to 3.0.0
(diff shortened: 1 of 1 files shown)


vincentsarago (approved):

CHANGES.md
```diff
@@ -6,6 +6,8 @@
 
 * add `grayscale` and `bitonal` algorithms
 * add `transform` and `crs` for `tiff` outputs
+* fix OpenAPI spec for `histogram_range` examples (@guillemc23, https://github.com/developmentseed/titiler/pull/1239)
```
vincentsarago:
```suggestion
```

PR #1248, opened 2025-10-23
fix: add opener_options arg to titiler.xarray.io.Reader

### LINKED ISSUE #1246: Add `opener_options` arg to titiler.xarray.io.Reader
#### Problem description

Users can provide their own `opener` function to the xarray `Reader` but sometimes it might be necessary to parameterize that function with additional args. Right now that is not possible.

I propose that we add an `opener_options` arg to `titiler.xarray.io.Reader` to make it possible to pass arguments through to the `opener`.

For example, there are cases where the storage credentials might need to be passed on a per-dataset basis so it would be helpful to be able to configure that via the `opener`.

Ref: https://github.com/developmentseed/titiler-cmr/issues/88

hrodmn:
This adds some flexibility to the custom `opener` option that some applications use by creating a mechanism to pass a set of arguments to the `opener` via `opener_args` a la https://github.com/EOPF-Explorer/titiler-eopf/blob/main/titiler/eopf/reader.py#L243.

This will help in cases like https://github.com/developmentseed/titiler-cmr/issues/88

resolves #1246

### DIFF
--- a/src/titiler/xarray/tests/test_io_tools.py
+++ b/src/titiler/xarray/tests/test_io_tools.py
@@ -285,7 +285,7 @@ def test_opener():
             opener_options={"special_arg": False},
             variable="dataset",
         ) as src:
-            assert src.info()
+            pass
 
 
 @pytest.mark.parametrize(


src/titiler/xarray/tests/test_io_tools.py
```diff
@@ -223,6 +225,69 @@ def test_reader(protocol, filename):
...
+        assert src.info()
+
+    with pytest.raises(ValueError):
+        with Reader(
+            src_path=src_path,
+            opener=custom_netcdf_opener,
+            opener_options={"special_arg": False},
+            variable="dataset",
+        ) as src:
+            assert src.info()
```
vincentsarago:
```suggestion
            pass
```

Issue #1084, opened 2025-01-28
Use obstore to access remote datasets in xarray_open_dataset


maxrjones:
@sharkinsspatial suggested in our ODD checkin yesterday trying out obstore in the XarrayReader. Since titiler doesn't rely on any distributed frameworks, I believe this should be possible today using the [fsspec integration](https://developmentseed.org/obstore/latest/api/fsspec/) to get a file handler for NetCDF/HDF files. I may try to run some quick benchmarks to see if this impacts performance in any way.

Once https://github.com/zarr-developers/zarr-python/pull/1661 is merged, obstore could be used for Zarr stores with [notable performance benefits](https://github.com/zarr-developers/zarr-python/pull/1661#issuecomment-2599299545).

vincentsarago:
Note: Obstore support in zarr-python was introduce in `zarr-python==3.0.7` but sadly zarr-python>=3.0 is only available for python >=3.11 https://github.com/zarr-developers/zarr-python/blob/v3.0.8/pyproject.toml#L33 😬 

we won't be able to support this until we remove support for python 3.9/3.10 in titiler.xarray

maxrjones:
> we won't be able to support this until we remove support for python 3.9/3.10 in titiler.xarray

Do you support python versions through their end of life? Zarr follows NEP 29 - https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table.

maxrjones:
> titiler.xarray requires xarray from rio-tiler and there is no version limit ([cogeotiff/rio-tiler@`deb5f3b`/pyproject.toml#L70-L73](https://github.com/cogeotiff/rio-tiler/blob/deb5f3b976a26b94677979a2f9c1b88d371446f0/pyproject.toml#L70-L73)) which means that we depends on versions that were compatible with python<3.11

titiler.xarray also separately declares xarray as a dependency:
https://github.com/developmentseed/titiler/blob/e4ca43483666628e49d48ccdafd42a93ffa74ba1/src/titiler/xarray/pyproject.toml#L32-L37

maxrjones:
ah sorry, I'm wasn't following that you allow people to use very old version of xarray, of course that makes sense

negin513:
Considering support for 3.9 and 3.10 has been dropped for [#180](https://github.com/developmentseed/morecantile/issues/180), do you think this might be a good moment to move the titiler-xarray base reader over to `obstore`?

vincentsarago:
[...]

PR #1253, opened 2025-10-29
switch titiler.xarray to obstore+zarr and add default application

vincentsarago:
ref https://github.com/developmentseed/titiler/discussions/1247
closes https://github.com/developmentseed/titiler/issues/1084

This PR does:
- create a new Opener function using Obstore and Zarr (only for ZARR Dataset) 
- switch the default reader to `open_zarr` opener
- add Zarr-python and Obstore as dependencies
- create a default application (for Zarr)
- remove python 3.10 support

### ToDo
- add tests for the default application
- update documentation 
- add example for application using fsspec opener

### DIFF
--- a/src/titiler/xarray/titiler/xarray/io.py
+++ b/src/titiler/xarray/titiler/xarray/io.py
@@ -32,7 +32,6 @@ def open_zarr(  # noqa: C901
     group: Optional[str] = None,
     decode_times: bool = True,
     decode_coords: str = "all",
-    consolidated: bool = True,
     infer_region: bool = True,
     **kwargs: Any,
 ) -> xarray.Dataset:
@@ -56,7 +55,6 @@ def open_zarr(  # noqa: C901
     # Default args
     xr_open_args: Dict[str, Any] = {
         "engine": "zarr",
-        "consolidated": consolidated,
         "decode_coords": decode_coords,
         "decode_times": decode_times,
     }
--- a/src/titiler/xarray/pyproject.toml
+++ b/src/titiler/xarray/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
     "xarray",
     "rioxarray",
     "obstore",
-    "zarr>=3.0,<4.0",
+    "zarr>=3.1,<4.0",
 ]
 
 [project.optional-dependencies]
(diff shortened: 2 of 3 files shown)


src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -27,64 +46,48 @@ def xarray_open_dataset(  # noqa: C901
...
-            }
-        )
-        fs = fsspec.filesystem(protocol)
-        ds = xarray.open_dataset(fs.open(src_path), **xr_open_args)
+    xr_open_args.update(
+        {
+            "engine": "zarr",
+            "consolidated": consolidated,
+        }
+    )
```
hrodmn: can we just add these to `xr_open_args` from the beginning?

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -238,3 +241,78 @@ def close(self):
...
+        xr_open_args.update(
+            {
+                "engine": "h5netcdf",
+                "lock": False,
+            }
+        )
+        fs = fsspec.filesystem(protocol, **kwargs)
+        ds = xarray.open_dataset(fs.open(src_path), **xr_open_args)
+
+    # Fallback to Zarr
```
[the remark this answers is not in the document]
vincentsarago: in theory you could but I think for compatibility issue it's better to keep everything separated.

> maybe it is nice to make this one reader that can open anything though.

User could do it if they need

maxrjones:
FYI this type of wrapper - https://github.com/virtual-zarr/obspec-utils/pull/1 - provides a way to use xarray + obstore with libraries that expect file handles (e.g., netcdf4/h5netcdf). That would help if the goal is just to limit dependencies, but wouldn't help if the goal is to simplify code complexity by removing support for data formats other than zarr.

vincentsarago:
are we good to go with this pr @hrodmn @maxrjones ?

src/titiler/xarray/titiler/xarray/io.py
```diff
@@ -1,20 +1,40 @@
...
 
-def xarray_open_dataset(  # noqa: C901
+
+@cache
+def open_zarr(  # noqa: C901
     src_path: str,
     group: Optional[str] = None,
     decode_times: bool = True,
+    decode_coords: str = "all",
+    consolidated: bool = True,
```
maxrjones: xarray will try to use consolidated metadata by default, and fall back to non-consolidated if it doesn't exist. By setting the default to consolidated=True, you will tell xarray to raise an error if consolidated metadata doesn't exist rather than falling back to distributed metadata. I'm not sure this restriction is worthwhile.

vincentsarago: 🙏 makes sense

PR #1254, opened 2025-10-29
Bump minimum supported python version to 3.11

maxrjones:
The python version should be specified as part of setup-uv rather than as a separate step, e.g., see workflow linked in https://github.com/developmentseed/titiler/pull/1245#discussion_r2473440497.


vincentsarago:
I think I'm in peace with updating the python version to 3.11 for all sub-modules 🙏

PR #1263, opened 2025-11-14
switch to `sel={dim}={method}::{value}` notation to specify selector method

vincentsarago:
match https://github.com/earth-mover/xpublish-tiles/pull/164 (🙏 @dcherian)

Before this PR we used `sel-method=nearest` parameter but this limited the possibility to use different method for different dimension.

With this PR we introduce a new string format for selectors: `sel={dimension}=({method}::)?{value}`


dcherian:
Haha I love that we're stealing ideas from each other. Hurray for open source!

abarciauskas-bgse:
Do we have a sense of pros and cons for each syntax option (i.e. `|`, `::`, others?) for selector parameters?

vincentsarago:
@dcherian can you share why you went with `::` as separator?

vincentsarago:
alright, I'm going to merge this as is, we could revisit for next major release

PR #1269, opened 2025-11-19
do not rename the TIME dimension

vincentsarago:


abarciauskas-bgse:
this LGTM ~did something happen to require this change?~

I see from another conversation this renaming was pulled over from another codebase and but not required here.

PR #1288, opened 2025-12-16
0.26.0 -> 1.0.0

vincentsarago:


deployment/k8s/charts/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: 0.26.0
+appVersion: 1.0.0
 description: A dynamic Web Map tile server
 name: titiler
-version: 1.2.14
+version: 1.2.15
```
ciaransweet: Probably 1.3/2 no?

Deploying a major version?

vincentsarago: well the chart itself didn't change 🤷

PR #1292, opened 2025-12-19
feat!: add titiler-xarray to helm chart by supporting top level titiler and titiler_xarray

ciaransweet:
## What I am changing

Closes: https://github.com/developmentseed/titiler/issues/1289

This PR adds `titiler-xarray` as an option to deploy via the helm chart.

Doing so I've made it so you can deploy `titiler`, `titiler-xarray` or both together.

## How I did it

- Added corresponding template additions for `titiler-xarray` in all templates in `deployment/k8s/charts/templates`
- Duplicated values entries for chart and added one more level to the structure to support `titiler` and `titiler-xarray`

## How you can test it

You can helm install this chart into your cluster, see below for an example of this working:

image

image

image

### DIFF
--- a/deployment/k8s/charts/templates/configmap.yaml
+++ b/deployment/k8s/charts/templates/configmap.yaml
@@ -8,8 +8,8 @@ data:
   netrc: {{ tpl (.Values.titiler.netrc) . | quote }}
   {{- end }}
 {{- end }}
----
 {{- if .Values.titiler_xarray.enabled }}
+---
 apiVersion: v1
 kind: ConfigMap
 metadata:
--- a/deployment/k8s/charts/templates/deployment.yaml
+++ b/deployment/k8s/charts/templates/deployment.yaml
@@ -100,8 +100,8 @@ spec:
         {{- toYaml . | nindent 8 }}
     {{- end }}
 {{- end}}
----
 {{- if .Values.titiler_xarray.enabled }}
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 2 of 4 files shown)


deployment/k8s/charts/values.yaml
```diff
@@ -1,88 +1,191 @@
...
-fullnameOverride: ""
-
-terminationGracePeriodSeconds: 30
-
-service:
-  type: ClusterIP
-  port: 80
-
-ingress:
+titiler:
```
emmanuelmathot: I don't like having titiler as a component of itself. The chart is about titiler so it should stay at the root.

deployment/k8s/charts/values.yaml
```diff
@@ -1,88 +1,191 @@
...
+
+  podSecurityContext: {}
+    # fsGroup: 1001
+    # runAsNonRoot: true
+    # runAsUser: 1001
+
+  netrc: ""   # File content of netrc
+
+# Default values for titiler-xarray.
+titiler_xarray:
```
emmanuelmathot: xarray should be an option through `xarray.enabled` and should reuse as much as possible the titiler core config. Only the config specific to xarray should be added.

emmanuelmathot (changes requested):
Sorry for being late on this one. In terms of deployment xarray should not be a separate deployment but included as an option in the core app or a simple switch with raster. I can understand this is not possible now but I want to keep the deployment interface for the user as a single titiler component with the xarray option.

deployment/k8s/charts/values.yaml
ciaransweet: Could yourself and @vincentsarago discuss these before tickets get created? It was said it should be able to be separate apps https://github.com/developmentseed/titiler/issues/1289#issuecomment-3664763316

If its not this, sounds like some application changes need to be made first?

PR #1313, opened 2026-02-06
1308 input validation changes

### LINKED ISSUE #1308: Bad requests can return HTTP 500 responses
#### Problem description

Some query string parameters accept string values, and expect specific string formats, but do not enforce input validation. This results in exceptions and HTTP 500 responses to requests that should receive HTTP 4xx responses.

Example using `rescale`:

```sh
curl -I -X GET 'http://localhost:8000/cog/tiles/WebMercatorQuad/0/0/0.png?url=http://some.cog/url&rescale=non-numeric-value'
HTTP/1.1 500 Internal Server Error
date: Wed, 04 Feb 2026 17:27:48 GMT
server: uvicorn
content-length: 67
content-type: application/json
```

Some other examples from initial testing that can return HTTP 500 responses:
- http://localhost:8000/cog/tiles/WebMercatorQuad/0/0/0?url=some.cog/url&algorithm_params=non-json-value (note that `algorithm_params` does have [some validation logic](https://github.com/developmentseed/titiler/blob/d4df94a6587a49af50068f035371eed00b8ae305/src/titiler/core/titiler/core/algorithm/__init__.py#L94) but this does not cover invalid JSON)
- http://localhost:8000/cog/tiles/WebMercatorQuad/0/0/0?url=some.cog/url&expression=invalid-value
- http://localhost:8000/cog/tiles/WebMercatorQuad/0/0/0?url=some.cog/url&nodata=invalid-value

This behaviour misinforms the API client, and does not communicate that modifying the request could fix the issue. Depending on the specifics it might also risk exposing internal implementation details to the client unnecessarily.

I plan to submit a PR to address these issues but I would first like to confirm that titiler maintainers agree that it should be addressed.

#### Expected Output

Expected output in this scenario would be an HTTP 422 (in line with other endpoints) or HTTP 400 response.

#### Environment Information

- macOS Sequoia
- Python 3.12.11
- `main` branch at [d4df94a6587a49af50068f035371eed00b8ae305](https://github.com/developmentseed/titiler/commit/d4df94a6587a49af50068f035371eed00b8ae305)

captaincoordinates:
Closes #1308 

Addresses several instances of validation failures returning HTTP 500 instead of 4xx responses, with appropriate testing. Problematic endpoints identified via a combination of manual code review and automated [CATS](https://endava.github.io/cats/) testing.

In some cases validation is possible with simple regular expressions as part of a `Query` definition, however in other cases either imperative logic is required or a regular expression would be so long or complex as to be onerous.

In a small number of cases it is only possible to validate input by attempting to process it and catching any exceptions. In-keeping with existing logic, any validation errors encountered after Query / dependency logic is complete are reported as HTTP 400 rather than 422.

### DIFF
--- a/src/titiler/core/titiler/core/factory.py
+++ b/src/titiler/core/titiler/core/factory.py
@@ -13,7 +13,7 @@ import jinja2
 import numpy
 import rasterio
 from attrs import define, field
-from fastapi import APIRouter, Body, Depends, HTTPException, Path, Query, status
+from fastapi import APIRouter, Body, Depends, Path, Query
 from fastapi.dependencies.utils import get_parameterless_sub_dependant
 from fastapi.params import Depends as DependsFunc
 from geojson_pydantic.features import Feature, FeatureCollection
@@ -24,7 +24,6 @@ from pydantic import Field
 from rio_tiler.colormap import ColorMaps
 from rio_tiler.colormap import cmap as default_cmap
 from rio_tiler.constants import WGS84_CRS
-from rio_tiler.errors import InvalidExpression
 from rio_tiler.io import BaseReader, MultiBandReader, MultiBaseReader, Reader
 from rio_tiler.models import ImageData, Info
 from rio_tiler.types import ColorMapType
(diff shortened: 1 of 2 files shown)


src/titiler/core/titiler/core/dependencies.py
```diff
@@ -578,6 +590,7 @@ class HistogramParams(DefaultDependency):
                     "value": "0,1000",
                 },
             },
+            pattern=separated_parseable_floats_regex(count=2),
```
vincentsarago: 👍

src/titiler/core/titiler/core/factory.py
```diff
@@ -914,15 +915,20 @@ def tile(
...
+                        image = src_dst.tile(
+                            x,
+                            y,
+                            z,
+                            tilesize=scale * 256,
+                            **tile_params.as_dict(),
+                            **layer_params.as_dict(),
+                            **dataset_params.as_dict(),
+                        )
+                    except InvalidExpression as e:
```
vincentsarago: this is going to be annoying to add this at each reader block. I think it would be better to add `InvalidExpression` in https://github.com/developmentseed/titiler/blob/main/src/titiler/core/titiler/core/errors.py#L33-L46

captaincoordinates: Good idea, again I applied this change and added you as co-author

PR #1334, opened 2026-03-09
fix: reset pixel_selection per feature in statistics FeatureCollection

raster-blaster:
## Bug

`POST /statistics` with a `FeatureCollection` containing 2+ distinct features returns **500 Internal Server Error**:

```
ValueError: operands could not be broadcast together with shapes (1,7,15) (1,4,11)
```

### Reproduction

| Test Case | Result |
|-----------|--------|
| Single Feature | 200 |
| FeatureCollection with 1 feature | 200 |
| Same feature duplicated in FC | 200 |
| **2 distinct features in FC** | **500** |
| **3 distinct features in FC** | **500** |

### Root Cause

In `MosaicTilerFactory.statistics()` → `geojson_statistics()`, the `pixel_selection` object is created once via FastAPI `Depends()` and reused across all features in the loop:

```python
for i, feature in enumerate(fc.features):
    image, assets = src_dst.feature(
        shape,
        pixel_selection=pixel_selection,  # <-- SAME instance reused!
        ...
    )
```

The `pixel_selection` (e.g., `FirstMethod`) stores `self.mosaic` — a numpy array with the shape of the first feature's raster data. When the second feature has a different geometry → different pixel dimensions → `numpy` broadcasting fails in `FirstMethod.feed()`:

```python
pidex = self.mosaic.mask & ~array.mask  # (1,7,15) & (1,4,11) = ValueError
```

Duplicating the same feature works because both produce identical array shapes.

### Fix

Create a fresh `pixel_selection` instance per feature in the loop:

```python
feature_pixel_selection = pixel_selection.__class__()
```

This ensures no stale mosaic state carries over between features.

### Tested with
- titiler-pgstac 1.3.0 + pgstac 0.9.1 (live Docker stack)
- titiler.mosaic 2.0.0b2 (source, new regression test added)
- New pytest: `test_mosaic_statistics_featurecollection` in `src/titiler/mosaic/tests/test_factory.py`

### DIFF
--- a/src/titiler/mosaic/titiler/mosaic/factory.py
+++ b/src/titiler/mosaic/titiler/mosaic/factory.py
@@ -75,7 +75,7 @@ def PixelSelectionParams(
     """
     Returns the mosaic method used to combine datasets together.
     """
-    return PixelSelectionMethod[pixel_selection].value()
+    return PixelSelectionMethod[pixel_selection].value
 
 
 def DatasetPathParams(url: Annotated[str, Query(description="Mosaic URL")]) -> str:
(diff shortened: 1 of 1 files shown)


src/titiler/mosaic/titiler/mosaic/factory.py
```diff
@@ -75,7 +75,7 @@ def PixelSelectionParams(
     """
     Returns the mosaic method used to combine datasets together.
     """
-    return PixelSelectionMethod[pixel_selection].value()
+    return PixelSelectionMethod[pixel_selection].value
```
vincentsarago: This is where the issue was because fastAPI is caching this. With the change, we pass the class which will then be initialized in the mosaic-reader

PR #1389, opened 2026-05-05
feat(mosaic): add zoom level params which override values from tiler backend.

underchemist:
## What I am changing

Adding zoom level dependency to MosaicTilerFactory, and use if set instead of tiler backend values.

## How I did it

Minimal change required to achieve expected behaviour of HTTP 400 response if MOSAIC_STRICT_ZOOM set and `minzoom/maxzoom` parameter passed to `/tiles` endpoint.

## How you can test it

Here is a repo which asserts expected behaviour https://github.com/underchemist/titiler-pgstac-zoom-override-example

## Related Issues

- related https://github.com/stac-utils/titiler-pgstac/issues/272

### DIFF
--- a/src/titiler/mosaic/titiler/mosaic/factory.py
+++ b/src/titiler/mosaic/titiler/mosaic/factory.py
@@ -3,6 +3,7 @@
 import logging
 import os
 from collections.abc import Callable
+from dataclasses import dataclass
 from typing import Annotated, Any, Literal
 from urllib.parse import urlencode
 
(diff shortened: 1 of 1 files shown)


vincentsarago:
@underchemist thanks for the PR 

IMO it feels weird to have to set min/max zoom to the tile endpoint. If the client know the min/max zoom range it shouldn't try to make a Z>maxzoom, Z<minzoom request 🤷

PR #1443, opened 2026-06-22
fix: only log non-generic 500 errors

### LINKED ISSUE #1440: Follow-up to #1429: avoid ERROR tracebacks for handled 4xx, and duplicate logging of the catch-all 500
Follow-up to #1429, which added `logger.error(..., exc_info=exc)` inside `exception_handler_factory`. Thanks @vincentsarago for turning #1424 around so quickly!

Now that the handler logs *every* mapped exception at `ERROR`, I think there are two cases worth separating before I send a PR with tests.

### 1. Several mapped exceptions are handled 4xx responses

`DEFAULT_STATUS_CODES` maps many cases to 4xx — e.g. `TileOutsideBounds` → 404, and `MissingBands` / `MissingAssets` / `InvalidColorFormat` / `InvalidExpression` → 400, `InvalidAssetName` / `InvalidBandName` → 404. These can occur routinely for public tile services (out-of-bounds tiles, bad params). Logging all of them at `ERROR` with `exc_info` produces tracebacks for what many deployments would treat as handled client errors rather than application failures, and adds log volume/cost (e.g. CloudWatch) for normal traffic.

### 2. The catch-all `Exception` (500) is already re-raised by Starlette

`DEFAULT_STATUS_CODES` includes `Exception: 500`. Starlette routes handlers registered for `500`/`Exception` to `ServerErrorMiddleware` (not `ExceptionMiddleware`); after calling the installed handler and sending the response, it deliberately re-raises so the server/adapter can log it:

```python
# starlette/middleware/errors.py
if not response_started:
    await response(scope, receive, send)
# We always continue to raise the exception.
# This allows servers to log the error ...
raise exc
```

So the server/adapter already logs the catch-all traceback — Mangum on Lambda does it via `mangum.protocols.http` (`logger.exception("An error occurred running the application.")`), and standard ASGI servers like uvicorn log the re-raised exception too. After #1429, titiler's own `logger.error(..., exc_info=exc)` duplicates that traceback for the catch-all path.

The genuinely uncovered case is the concrete 5xx mappings — `RasterioIOError`, `RasterioError`, `RioTilerError` — which go through `ExceptionMiddleware` and are *not* re-raised once the handler returns a response, so neither uvicorn nor Mangum ever logs them. (This is the distinction I tried to draw in #1424; logging those is the real win.)

**Minor:** the message uses an f-string rather than lazy `%s` args (`ruff` `G004` / `pylint` `W1203`).

### Proposed direction (happy to adjust)

- Don't emit `ERROR` + traceback for handled 4xx (skip, or log at `INFO`/`WARNING` without `exc_info`).
- Log the concrete mapped 5xx exceptions with `exc_info`.
- Exclude the catch-all `Exception`/500 from titiler's own traceback log, since `ServerErrorMiddleware` already routes it to the server/adapter.
- Use lazy `%s` formatting, and optionally `extra={"status_code": status_code}`; leave broader structured logging to #1190.

Concretely, something like:

```python
def exception_handler_factory(status_code: int, *, log_exception: bool = False) -> Callable:
    def handler(request: Request, exc: Exception):
        if status_code == status.HTTP_204_NO_CONTENT:
            return Response(content=None, status_code=204)
        if log_exception:
            logger.error(
                "Exception mapped to HTTP %s response",
                status_code,
                exc_info=exc,
                extra={"status_code": status_code},
            )
        return JSONResponse(content={"detail": str(exc)}, status_code=status_code)
    return handler

# ...
for exc, code in status_codes.items():
    app.add_exception_handler(
        exc,
        exception_handler_factory(code, log_exception=code >= 500 and exc is not Exception),
    )
```

This targets the real blind spot (concrete 5xx handled by `ExceptionMiddleware`) while avoiding 4xx traceback noise and catch-all duplication. Happy to open the PR with tests if this sounds reasonable.

vincentsarago:
closes https://github.com/developmentseed/titiler/issues/1440

@wakame1367 does this conform with the issue?

### DIFF
--- a/src/titiler/core/titiler/core/errors.py
+++ b/src/titiler/core/titiler/core/errors.py
@@ -51,7 +51,7 @@ DEFAULT_STATUS_CODES = {
 }
 
 
-def exception_handler_factory(status_code: int) -> Callable:
+def exception_handler_factory(status_code: int, exc_class: type[Exception]) -> Callable:
     """
     Create a FastAPI exception handler from a status code.
     """
@@ -60,9 +60,9 @@ def exception_handler_factory(status_code: int) -> Callable:
         if status_code == status.HTTP_204_NO_CONTENT:
             return Response(content=None, status_code=204)
 
-        # Only log >=500 errors but not generic exceptions (Exception)
-        # already logged by Starlette
-        if status_code >= 500 and exc is not Exception:
+        # Only log concrete >=500 errors; the catch-all Exception/500 is
+        # already re-raised and logged by Starlette's ServerErrorMiddleware.
+        if status_code >= 500 and exc_class is not Exception:
             logger.error(
                 "Exception mapped to HTTP %s response",
                 status_code,
@@ -81,4 +81,4 @@ def add_exception_handlers(
     Add exception handlers to the FastAPI app.
     """
     for exc, code in status_codes.items():
-        app.add_exception_handler(exc, exception_handler_factory(code))
+        app.add_exception_handler(exc, exception_handler_factory(code, exc))
--- a/src/titiler/core/titiler/core/errors.py
+++ b/src/titiler/core/titiler/core/errors.py
@@ -60,10 +60,14 @@ def exception_handler_factory(status_code: int) -> Callable:
         if status_code == status.HTTP_204_NO_CONTENT:
             return Response(content=None, status_code=204)
 
-        logger.error(
-            f"Exception mapped to HTTP {status_code} response",
-            exc_info=exc,
-        )
+        # Only log >=500 errors but not generic exceptions (Exception)
+        # already logged by Starlette
+        if status_code >= 500 and exc is not Exception:
+            logger.error(
+                "Exception mapped to HTTP %s response",
+                status_code,
+                exc_info=exc,
+            )
 
         return JSONResponse(content={"detail": str(exc)}, status_code=status_code)
 


wakame1367:
Thanks for putting this together @vincentsarago! The 4xx gate and the lazy `%s` formatting are exactly right. One thing on the catch-all suppression though — I don't think `exc is not Exception` does what the comment says:

Inside `handler(request, exc)`, `exc` is the **raised exception instance**, not the registered class. An instance is never identical to the `Exception` class object, so `exc is not Exception` is **always `True`** — including for the `Exception: 500` catch-all. So the catch-all still gets logged here, and `ServerErrorMiddleware` re-raises it so uvicorn/Mangum logs the same traceback again. That's the duplicate we were trying to drop.

Quick repro:

```python
>>> class RioTilerError(Exception): ...
>>> ValueError("x") is not Exception      # catch-all path
True
>>> Exception("x") is not Exception       # even a bare Exception instance
True
>>> RioTilerError("x") is not Exception   # concrete 5xx
True
```

The discriminator has to come from the **registered class**, which is only known at registration time. Threading it through the factory keeps it param-free at the call boundary:

```python
def exception_handler_factory(status_code: int, exc_class: type[Exception]) -> Callable:
    def handler(request: Request, exc: Exception):
        if status_code == status.HTTP_204_NO_CONTENT:
            return Response(content=None, status_code=204)

        # Only log concrete >=500 errors; the catch-all Exception/500 is
        # already re-raised and logged by Starlette's ServerErrorMiddleware.
        if status_code >= 500 and exc_class is not Exception:
            logger.error(
                "Exception mapped to HTTP %s response",
                status_code,
                exc_info=exc,
            )

        return JSONResponse(content={"detail": str(exc)}, status_code=status_code)

    return handler


def add_exception_handlers(app: FastAPI, status_codes: dict[type[Exception], int]) -> None:
    for exc, code in status_codes.items():
        app.add_exception_handler(exc, exception_handler_factory(code, exc))
```

Now `exc_class is not Exception` correctly skips only the catch-all, while `RasterioIOError` / `RasterioError` / `RioTilerError` still log with `exc_info`.

One caveat worth noting: an unmapped error (e.g. a stray `KeyError`) routes to the `Exception: 500` handler via MRO and would correctly be skipped here too — which is what we want, since those are exactly the ones Starlette already logs.

Happy to push this as a commit/PR if that's easier — just let me know.

vincentsarago:
👍 thanks for the explanation, I'll fix this

PR #1448, opened 2026-07-07
fix: fall back to WKT for WMTS BoundingBox crs when CRS cannot be resolved to a URN

### LINKED ISSUE #1043: raise warning/error or use WKT for WMTS boundingbox CRS when no URI exists
So an issue I've known would be coming my way is in the use of `CRS_to_uri` inside the WMTSCapabilities.xml endpoints for the bounds object CRS parameter, which when using a slightly tweaked/custom CRS can result in a `None` being specified in the WMTS XML document. This causes issues with clients like QGIS, but will work as soon as a valid CRS URI (or interestingly WKT) is swapped in.

After looking at this, I think I must conclude that this isn't a bug in pyproj/rasterio/rio-tiler. It is totally possible to define a tilematrixset or COG that uses a CRS that may use a geographic CRS that won't resolve with `to_authority` despite only minor modifications. 

Lowering the confidence parameter used in rio-tiler could help in some cases but won't work universally and could cause some other unintended issues. So it made the most sense to open an issue here and discuss solutions.

I think then there are 2/3 options

1) raise an error when `CRS_to_uri` evals to `None`  
2) raise a warning in the logs when this occurs and continue to allow `None` to pass through
3) Fall back to the WKT 

After reviewing OGC 06-121r9 (sections 10.2-10.3) it does look like for the most part clients would be expecting this CRS parameter to be a URI/URN. However, while testing I was able to use the full WKT string for the crs parameter and QGIS/GDAL was able to use it correctly. Given the ubiquity of GDAL, my expectation would be that this could be seen as a sensible fall back even if it's technically out-of-spec with the OGC, although maybe this is just plainly a bad idea... 

My suspicion then is that option 1 would also be too aggressive of a fix and that it might just be best to keep the change in behavior minimal by logging a warning. 

I am happy to implement this, but wanted to check with others before starting to see if there are other ideas or ways to work with the WMTS spec in ways I hadn't considered

wakame1367:
closes #1043

## What

When the `wmtsExtension` is configured with a custom CRS that cannot be
resolved to an authority identifier (e.g. a slightly tweaked polar
stereographic projection), `CRS_to_urn()` returns `None` and the WMTS
Capabilities document ends up with an invalid bounding box CRS:

```xml
<ows:BoundingBox crs="None">
```

This breaks WMTS clients such as QGIS.

This PR implements the fallback discussed in #1043: when no URN can be
resolved, we now emit a `UserWarning` and use the WKT representation of the CRS
for the `crs` attribute. WKT is not strictly compliant with the WMTS 1.0.0
expectation that this value be a CRS URI/URN, but GDAL/QGIS handle it
correctly, and it is preferable to emitting `crs="None"`.

The same fix is applied to both `titiler.extensions.wmtsExtension` and
`titiler.mosaic.extensions.wmtsExtension`, and the now-unneeded
`# type: ignore` comments are removed.

## Changes

- `titiler/extensions/wmts.py` / `titiler/mosaic/extensions/wmts.py`: fall back
  to `geographic_crs.to_wkt()` (with a `UserWarning`) when
  `CRS_to_urn(geographic_crs)` returns `None`
- add tests for both extensions using a custom CRS that cannot be resolved to
  an authority, asserting that a `UserWarning` is emitted and that the `crs`
  attribute of `ows:BoundingBox` contains the WKT string
- update `CHANGES.md`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+### titiler.extensions
+
+* fall back to WKT (with a warning) for the WMTS `BoundingBox` crs attribute when the CRS cannot be resolved to a URN (https://github.com/developmentseed/titiler/issues/1043)
+
+### titiler.mosaic
+
+* fall back to WKT (with a warning) for the WMTS `BoundingBox` crs attribute when the CRS cannot be resolved to a URN (https://github.com/developmentseed/titiler/issues/1043)
+
 ## 2.0.5 (2026-06-18)
 
 ## What's Changed
--- a/src/titiler/extensions/tests/test_wmts.py
+++ b/src/titiler/extensions/tests/test_wmts.py
@@ -2,10 +2,14 @@
 
 import io
 import os
+import xml.etree.ElementTree as ET
 
+import pytest
 import rasterio
 from fastapi import FastAPI
 from owslib.wmts import WebMapTileService
+from rasterio.crs import CRS
+from rio_tiler.utils import CRS_to_urn
 from starlette.testclient import TestClient
 
 from titiler.core.factory import TilerFactory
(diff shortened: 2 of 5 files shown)


wakame1367:
The remaining red on this PR's `tests` jobs is **not** caused by this change.
The mypy step fails on two pre-existing `[call-overload]` errors in
`titiler/core/algorithm/dem.py` (lines 182 and 242) that are already failing on
`main` — see e.g. the CI run for the latest merge commit on `main`, which fails
with the same two errors.

I opened #1449 to fix that pre-existing mypy failure separately. Once it lands
(and this branch is rebased on top), this PR's CI should be green.

vincentsarago:
thanks @wakame1367 🙏

PR #1459, opened 2026-07-22
feat(#1458): control XML layer identifiers

### LINKED ISSUE #1458: Ability to control layer identifiers in WMTS XML
I would like the ability to control layer identifiers in WMTS XML documents. I have a PR that I will submit after creating this issue. This is a feature request but it is so small that I do not think it warrants a discussion - please let me know if I am incorrect.

Currently [this line](https://github.com/developmentseed/titiler/blob/044c5b37fce6a69672e8a3d2aa53ad9916ede691/src/titiler/extensions/titiler/extensions/wmts.py#L193) defines layer identifiers (there is a very similar mosaic equivalent):

```python
title = src_path if isinstance(src_path, str) else "TiTiler"
```

This can create awkward and long layer identifiers that are difficult for some WMTS clients to work with. I propose an optional argument with the ability to control the resulting string, for example:

```python
# Called with a single argument `src_path` if not None.
# `src_data` is not guaranteed to be a str.
layer_identifier_provider: Callable[[Any], str] | None = field(default=None)
```

captaincoordinates:
Closes #1458 

I added a test for titiler.extentions.wmts, however I did not add one for titiler.mosaic.extentions.wmts as there are no existing tests for this module.

The following commands execute successfully:

```sh
uv run pytest src/titiler/extensions --cov=titiler.extensions --cov-report=term-missing && uv run pytest src/titiler/mosaic --cov=titiler.mosaic --cov-report=term-missing
```
```sh
uv run pre-commit run --all-files
```

### DIFF
--- a/src/titiler/mosaic/titiler/mosaic/extensions/wmts.py
+++ b/src/titiler/mosaic/titiler/mosaic/extensions/wmts.py
@@ -55,7 +55,9 @@ class wmtsExtension(FactoryExtension):
     # Note: Those dependencies should only require Query() inputs
     tile_dependencies: list[Callable] | None = field(default=None)
 
-    layer_identifier_provider: Callable[[Any], str] | None = field(default=None)
+    layer_identifier_provider: Callable[[Any], str] = field(
+        default=lambda x: x if isinstance(x, str) else "TiTiler Mosaic"
+    )
 
     # TODO: Remove in 3.0
     def __attrs_post_init__(self):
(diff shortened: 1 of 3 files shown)


src/titiler/mosaic/titiler/mosaic/extensions/wmts.py
```diff
@@ -55,6 +55,8 @@ class wmtsExtension(FactoryExtension):
     # Note: Those dependencies should only require Query() inputs
     tile_dependencies: list[Callable] | None = field(default=None)
 
+    layer_identifier_provider: Callable[[Any], str] | None = field(default=None)
```
vincentsarago: what do you think about adding `lambda: src_path if isinstance(src_path, str) else "TiTiler Mosaic"` as default for this so we avoid the `if/else` block

PR #1476, opened 2026-08-04
docs(security): document the GDAL 3.12 VRT options and how to remove the expression parameter

ishiland:
Two documentation updates to `SECURITY.md`, from running TiTiler as an internal sidecar behind an authorizing proxy.

### GDAL 3.12 VRT options

The page currently says new variables "might be introduced" in GDAL 3.12 and links the PR. They shipped, so this names them and their defaults:

- `GDAL_VRT_ENABLE_RAWRASTERBAND`, default `YES`
- `GDAL_VRT_RAWRASTERBAND_ALLOWED_SOURCE`, default `SIBLING_OR_CHILD_OF_VRT_PATH`

Both defaults are verified against the [VRT driver docs](https://gdal.org/en/stable/drivers/raster/vrt.html). This matters for deployments where `GDAL_SKIP=VRT` is not an option because VRT datasets are a product feature, which the existing text does not cover.

I also added the GDAL 3.13.2 restriction on the `/vsicurl/` `header_file` option, verified against the [virtual file systems docs](https://gdal.org/en/stable/user/virtual_file_systems.html).

### Removing the `expression` parameter

The page documents restricting sources with `path_dependency` but nothing about band math. Applications that only need band selection can drop the parameter entirely with `layer_dependency=BidxParams`, which is a stronger guarantee than relying on validation.

I ran the example against `titiler.core` 2.2.1 and confirmed `expression` disappears from the OpenAPI schema while `bidx` is unaffected.

Happy to split this into two PRs, drop either half, or move the band math section elsewhere if the security page is not the right home.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -33,7 +33,16 @@ see https://gdal.org/en/stable/user/security.html#gdal-vrt-driver
 
 Thus we recommend deploying titiler in infrastructure with limited access to the filesystem. Users can also `disable` the VRT driver completely by using `GDAL_SKIP=VRT` environment variable.
 
-In GDAL 3.12, new environment variables might be introduced to enable more control over the VRT driver: https://github.com/OSGeo/gdal/pull/12669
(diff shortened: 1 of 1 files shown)


vincentsarago (approved):
😍

PR #1496, opened 2026-09-11
fix(docker): keep pip in the application virtualenv

### LINKED ISSUE #1495: Titiler:2.2.2 pip install in downstream Dockerfiles installs into wrong environment (venv has no pip)
#### Problem description

The `titiler:2.2.2` image breaks subsequent `pip install` commands in downstream Dockerfiles.

I ran into a `ModuleNotFoundError` in my container, despite a `RUN pip install ...` command that previously worked fine against an older image tag.

`2.2.2` runs the app from a virtualenv at `/opt/venv`, created by [`uv`](https://github.com/astral-sh/uv) (confirmed via `/opt/venv/pyvenv.cfg`, which lists `uv = 0.12.11`). `uv`-created venvs commonly ship without a `pip` binary or module, by design — `uv` doesn't need pip to install packages.

Meanwhile, the `pip` resolved on `PATH` is a separate, `/usr/local`-based install, entirely disconnected from `/opt/venv`.

The result: any `RUN pip install ...` in a downstream Dockerfile installs into `/usr/local`, but the app's actual runtime (`/opt/venv/bin/python`, per `CMD`) only sees `/opt/venv`. Packages get installed somewhere the app never looks.

#### Reproduce

**Confirming the split:**

```bash
docker run --rm ghcr.io/developmentseed/titiler:2.2.2 sh -c \
  "which python; which pip"
# /opt/venv/bin/python
# /usr/local/bin/pip
```

**Confirming pip is absent from the venv itself:**

```bash
docker run --rm ghcr.io/developmentseed/titiler:2.2.2 sh -c \
  "ls /opt/venv/bin | grep -i pip; /opt/venv/bin/python -m pip --version"
# (no pip listed)
# /opt/venv/bin/python: No module named pip
```

**Confirming the venv is uv-managed:**

```bash
docker run --rm ghcr.io/developmentseed/titiler:2.2.2 cat /opt/venv/pyvenv.cfg
# uv = 0.12.11
# include-system-site-packages = false
```

**Full end-to-end repro (build + run, showing the failure):**

```bash
docker build -t titiler-repro - <<'EOF'
FROM ghcr.io/developmentseed/titiler:2.2.2
RUN pip install --no-cache-dir mangum
CMD ["python", "-c", "import mangum"]
EOF

docker run --rm titiler-repro
# Traceback (most recent call last):
#   ...
# ModuleNotFoundError: No module named 'mangum'
```

#### Environment

Docker on WSL2 (Ubuntu distro).

- Base image: `ghcr.io/developmentseed/titiler:2.2.2`
- Digest: `sha256:d073928eb0d2499c80ead6efb3639817819f946e17f691075fab6f554b27c4e2`

(Note: I originally hit this against `:latest`, which resolved to the same `titiler-application` version at the time. I'm pinning to `2.2.2` explicitly here since `:latest` can move while this issue is open.)

#### Expected behavior

A `pip install` in a `RUN` layer installs packages that are visible to the image's own entrypoint. This works correctly on `titiler:2.2.1`.

#### Workaround
Pinnig to `titiler:2.2.1` resolves the issue, so this appears to be specific only to the latest image.

mikemikimike:
## Summary

The 2.2.2 image runs the application from `/opt/venv`, but the uv-created
virtual environment does not contain `pip`. As a result, a downstream
Dockerfile's bare `pip install` resolves to the system interpreter instead of
the environment used by TiTiler.

## Changes

- Install `pip` into `/opt/venv` during the builder stage.
- Make the copied virtual environment writable by the existing non-root runtime user.
- Keep the existing runtime PATH and application dependency layout unchanged.

## Issue

Closes #1495

## Verification

- `uv sync --frozen` and the project mypy command in the official `codex-ci/python:3.14-build` pre-PR runner — PASS (54 source files, task `prepr-20260911T045200Z-15993`).
- Fixed downstream-container simulation — PASS: as `uid=1000(user)`, bare `pip` resolves to `/opt/venv/bin/pip`, `pip install mangum` succeeds, and `/opt/venv/bin/python` imports it from `/opt/venv/lib/python3.14/site-packages`.
- Core, xarray, and mosaic test modules in the same runner — PASS (92, 59, and 10 tests respectively).
- Application tests in the same runner — 39 passed, 1 failed in an existing Brotli/httpx2 compatibility path (`Decompressor.decompress()` does not accept `output_buffer_limit`); the failure is outside this Dockerfile-only change.
- `docker build --file dockerfiles/Dockerfile ...` — BLOCKED: a retry reached the Dockerfile but failed at the existing runtime `apt-get update` layer because the configured mirror returned a connection failure/404 for Debian trixie `Packages`, before the new `pip`/`chown` layers were evaluated. An earlier attempt failed while resolving `python:3.14`.

## Known limitations

- The local Docker image build is blocked by the configured Docker/Debian mirrors. The hosted pull-request workflows currently require repository-maintainer approval, and the repository's Docker publish workflow is not configured as a pull-request check.
- The pre-PR runner's `pre-commit` invocation could not use its generated Git snapshot, so that hook was not independently verified there.

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -60,7 +60,8 @@ RUN apt-get update && \
 COPY --from=builder /opt/venv /opt/venv
 
 RUN groupadd -g 1000 user && \
-    useradd -u 1000 -g user -s /bin/bash -m user
+    useradd -u 1000 -g user -s /bin/bash -m user && \
+    chown -R user:user /opt/venv
 
 USER user
 
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -33,6 +33,9 @@ COPY pyproject.toml uv.lock README.md LICENSE ./
 COPY src/titiler/ ./src/titiler/
 RUN uv sync --frozen --no-dev --no-editable --group server
 RUN uv pip install --no-deps --no-editable .
+# Keep pip inside the runtime venv so downstream images install into the
+# environment used by the application rather than the system interpreter.
+RUN uv pip install --python /opt/venv/bin/python pip
 
 # Runtime stage
 FROM python:${PYTHON_VERSION}-slim


vincentsarago:
closing for now until PR respect our AI policy

